PR #60, opened 2026-04-17
Adopt shared forbidden responses in customer portal

mihaipxm:
## Summary
- route customer portal companion ownership failures through the shared Hono API error serializer
- reuse `ForbiddenApiError` instead of package-local 403 payload construction
- keep the customer portal auth/transport surface aligned with the other route guard cleanup slices

## Validation
- pnpm -C packages/customer-portal test
- pnpm -C packages/customer-portal typecheck
- pnpm -C packages/customer-portal build
- git diff --check


mihaipxm:

PR #68, opened 2026-04-17
Add auth and identity architecture guide

mihaipxm:
## Summary
- add a concrete auth and identity guide for shared session auth, actor context, and permission boundaries
- document how templates and modules should consume the shared auth surface
- link the guide from the platform surface roadmap

## Testing
- git diff --check


mihaipxm:

PR #80, opened 2026-04-17
Reduce plugin-first package docs

mihaipxm:
## Summary
- tighten Hono and core package docs so plugins are framed as optional distribution bundles
- update sync plugin READMEs to present their adapter/subscriber role first and plugin usage second
- keep package examples aligned with the module/provider/extension/plugin taxonomy

## Testing
- git diff --check


mihaipxm:

PR #88, opened 2026-04-17
Use notification task runtime in workers

mihaipxm:
## Summary
- move operator, dmc, and dev worker entrypoints onto the shared notifications task runtime
- keep template-specific provider selection, but stop reconstructing providers ad hoc at each reminder task call
- align cron/worker notification dispatch with the shared package runtime seam

## Testing
- git diff --check
- pnpm test


mihaipxm:

PR #92, opened 2026-04-17
Adopt admin widget extension surface

mihaipxm:
## Summary
- add a real operator dashboard widget through the admin extension registry
- keep the widget contribution template-local so the extension surface is exercised without adding a dynamic plugin runtime
- use the existing dashboard metrics widget slot instead of duplicating page wiring

## Testing
- git diff --check
- pnpm -C templates/operator typecheck
- pnpm test


mihaipxm:

PR #95, opened 2026-04-17
Use shared route parsers in public product routes

mihaipxm:
## Summary
- replace manual query parsing in public product routes with the shared Hono parser helper
- keep the public products surface aligned with the shared route-authoring conventions
- leave route behavior unchanged while removing repeated URLSearchParams parsing

## Testing
- pnpm -C packages/products typecheck
- pnpm -C packages/products test
- git push (full repo pre-push suite)


mihaipxm:

PR #97, opened 2026-04-17
Use shared route parsers in public finance routes

mihaipxm:
## Summary
- replace manual query/body parsing in public finance routes with the shared Hono parser helpers
- keep the public finance surface aligned with the shared route-authoring conventions
- leave route behavior unchanged while removing repeated request parsing code

## Testing
- pnpm -C packages/finance typecheck
- pnpm -C packages/finance test
- git push (full repo pre-push suite)


mihaipxm:

PR #101, opened 2026-04-17
Use shared route parsers in checkout routes

mihaipxm:
## Summary
- replace direct JSON parsing in checkout collection routes with the shared Hono body parser
- replace manual query parsing in checkout admin reminder routes with the shared Hono query parser
- keep the preview endpoint's optional-empty-body behavior unchanged

## Testing
- pnpm -C packages/checkout typecheck
- pnpm -C packages/checkout test


mihaipxm:

PR #106, opened 2026-04-17
Use shared route parsers in transaction routes

mihaipxm:
## Summary
- replace manual query parsing in transaction offer and order routes with the shared Hono query parser
- replace repeated direct JSON parsing in transaction routes with the shared Hono body parser
- keep the transactions route surface aligned with the broader transport cleanup

## Testing
- pnpm -C packages/transactions typecheck
- pnpm -C packages/transactions test


mihaipxm:

PR #110, opened 2026-04-17
Use shared route parsers in pricing rules routes

mihaipxm:
## Summary
- replace manual query parsing in pricing rules routes with the shared Hono query parser
- replace repeated direct JSON parsing in pricing rules write routes with the shared Hono body parser
- keep the pricing rules transport surface aligned with the broader route cleanup

## Testing
- pnpm -C packages/pricing typecheck
- pnpm -C packages/pricing test


mihaipxm:

PR #111, opened 2026-04-17
Use shared route parsers in pricing core routes

mihaipxm:
## Summary
- replace manual query parsing in pricing core routes with the shared Hono query parser
- replace repeated direct JSON parsing in pricing core write routes with the shared Hono body parser
- keep the pricing core transport surface aligned with the broader route cleanup

## Testing
- pnpm -C packages/pricing typecheck
- pnpm -C packages/pricing test


mihaipxm:

PR #115, opened 2026-04-17
Use shared route parsers in ground routes

mihaipxm:
## Summary
- replace manual query parsing in ground routes with the shared Hono query parser
- replace repeated direct JSON parsing in ground create and update routes with the shared Hono body parser
- keep the ground transport surface aligned with the broader route cleanup

## Testing
- pnpm -C packages/ground lint
- pnpm -C packages/ground typecheck
- pnpm -C packages/ground test


mihaipxm:

PR #118, opened 2026-04-17
Use shared route parsers in resources routes

mihaipxm:
## Summary
- replace manual resource route query parsing with the shared Hono query parser
- replace direct JSON body parsing in resource and batch mutation endpoints with the shared Hono body parser
- keep the resources admin surface aligned with the bounded transport cleanup work

## Testing
- git diff --check
- pnpm -C packages/resources lint
- pnpm -C packages/resources typecheck
- pnpm -C packages/resources test


mihaipxm:

PR #121, opened 2026-04-17
Use shared route parsers in distribution routes

mihaipxm:
## Summary
- replace manual query parsing in distribution routes with the shared Hono query parser
- replace direct JSON body parsing in distribution create, update, and batch mutation endpoints with the shared Hono body parser
- keep the distribution transport surface aligned with the rest of the route-helper adoption work

## Testing
- git diff --check
- pnpm -C packages/distribution lint
- pnpm -C packages/distribution typecheck
- pnpm -C packages/distribution test


mihaipxm:

PR #125, opened 2026-04-17
Use shared route parsers in legal contract routes

mihaipxm:
## Summary
- replace manual contract route query parsing with the shared Hono query parser
- replace direct JSON body parsing in standard legal contract endpoints with the shared Hono body parser
- keep the optional empty-body document generation endpoints unchanged while aligning the rest of the transport surface

## Testing
- git diff --check
- pnpm -C packages/legal lint
- pnpm -C packages/legal test


mihaipxm:

PR #133, opened 2026-04-17
Use shared route parsers in extras routes

mihaipxm:
## Summary
- switch extras route query parsing to the shared Hono query parser
- switch extras route JSON body parsing to the shared Hono body parser
- keep the transport cleanup moving through small bounded route files

## Testing
- git diff --check
- pnpm -C packages/extras lint
- pnpm -C packages/extras typecheck
- pnpm -C packages/extras test


mihaipxm:

PR #134, opened 2026-04-17
Use shared route parsers in supplier routes

mihaipxm:
## Summary
- switch supplier route query parsing to the shared Hono query parser
- switch supplier route JSON body parsing to the shared Hono body parser
- keep the transport cleanup moving through bounded route files

## Testing
- git diff --check
- pnpm -C packages/suppliers lint
- pnpm -C packages/suppliers typecheck
- pnpm -C packages/suppliers test


mihaipxm:

PR #135, opened 2026-04-17
Use shared route parsers in booking requirements routes

mihaipxm:
## Summary\n- replace manual query parsing in booking requirements routes with the shared Hono query parser\n- replace direct JSON body parsing in create/update routes with the shared Hono body parser\n- keep the route surface behavior unchanged while aligning it with the transport conventions\n\n## Testing\n- git diff --check\n- pnpm -C packages/booking-requirements lint\n- pnpm -C packages/booking-requirements typecheck\n- pnpm -C packages/booking-requirements test


mihaipxm:

PR #136, opened 2026-04-17
Await shared query parsing in market routes

mihaipxm:
## Summary\n- await the shared Hono query parser in market route list endpoints\n- keep the existing route surface and schemas unchanged\n- align market routes with the rest of the shared transport cleanup\n\n## Testing\n- git diff --check\n- pnpm -C packages/markets lint\n- pnpm -C packages/markets typecheck\n- pnpm -C packages/markets test


mihaipxm:

PR #137, opened 2026-04-17
Use shared route parsers in identity routes

mihaipxm:
## Summary\n- replace manual query parsing in identity routes with the shared Hono query parser\n- replace direct JSON body parsing in create/update routes with the shared Hono body parser\n- keep the route surface unchanged while aligning it with the transport conventions\n\n## Testing\n- git diff --check\n- pnpm -C packages/identity lint\n- pnpm -C packages/identity typecheck\n- pnpm -C packages/identity test


mihaipxm:

PR #138, opened 2026-04-17
Use shared route parsers in policy routes

mihaipxm:
## Summary
- replace manual query parsing in policy routes with the shared Hono query parser
- replace direct JSON body parsing in policy admin/public routes with the shared Hono body parser
- keep route behavior unchanged while aligning it with the transport conventions

## Testing
- git diff --check
- pnpm -C packages/legal lint
- pnpm -C packages/legal test


mihaipxm:

PR #139, opened 2026-04-17
Use shared route parsers in contract routes

mihaipxm:
## Summary
- replace manual query parsing in contract admin/public routes with the shared Hono query parser
- replace direct JSON body parsing in contract admin/public routes with the shared Hono body parser
- keep optional empty-body document generation endpoints unchanged while aligning the rest of the route surface with the transport conventions

## Testing
- git diff --check
- pnpm -C packages/legal lint
- pnpm -C packages/legal test


mihaipxm:

PR #144, opened 2026-04-17
Reframe package surfaces around modules and providers

mihaipxm:
## Summary
- reframe `@voyantjs/core` around framework primitives first and plugin bundles second
- reframe `@voyantjs/hono` around modules, extensions, and admin/public route surfaces instead of plugin-led composition
- tighten `@voyantjs/checkout` wording so provider adapters come first and plugin bundles remain optional distribution wrappers

## Testing
- `git diff --check`


mihaipxm:

PR #145, opened 2026-04-17
Reframe plugin package readmes around adapters

mihaipxm:
## Summary
- lead the Payload, Sanity, SmartBill, and Netopia package READMEs with their adapter/subscriber role instead of plugin-first wording
- describe the plugin bundle as the distribution wrapper on top of those integrations
- tighten the exports tables so `./plugin` is clearly a bundle factory, not the primary architecture concept

## Testing
- `git diff --check`


mihaipxm:

PR #146, opened 2026-04-17
Note Hono bundle taxonomy follow-up

mihaipxm:
## Summary
- record the Hono transport naming follow-up in the platform roadmap
- clarify in the Hono README that @voyantjs/hono is the default transport adapter, not an optional integration plugin
- tighten the HonoPlugin source docs so it reads as the current transport-specific bundle contribution surface

## Testing
- git diff --check


mihaipxm:

PR #147, opened 2026-04-17
Use shared optional body parser in dev auth handler

mihaipxm:
## Summary
- replace the manual optional JSON parse in the dev auth workspace switch handler
- use the shared `parseOptionalJsonBody` helper from `@voyantjs/hono`
- keep the route behavior unchanged while aligning it with the shared transport helpers

## Testing
- git diff --check
- pnpm -C apps/dev typecheck


mihaipxm:

PR #149, opened 2026-04-17
Use shared route parsers in CRM custom field routes

mihaipxm:
## Summary
- replace manual query parsing in CRM custom-field definition and value routes
- replace manual JSON body parsing in the matching create, update, and upsert endpoints
- keep the CRM route behavior unchanged while aligning it with the shared Hono transport helpers

## Testing
- git diff --check
- pnpm -C packages/crm lint
- pnpm -C packages/crm typecheck
- pnpm -C packages/crm test


mihaipxm:

PR #156, opened 2026-04-18
Add SmartBill sync runtime helper

mihaipxm:
## Summary
- extract SmartBill client/logger/mapper/event-name assembly into a shared sync runtime helper
- refactor the SmartBill plugin to build subscribers from that runtime instead of inlining setup
- add focused unit coverage for the new helper while keeping plugin behavior the same

## Testing
- git diff --check
- pnpm -C packages/plugins/smartbill lint
- pnpm -C packages/plugins/smartbill typecheck
- pnpm -C packages/plugins/smartbill test


mihaipxm:

PR #157, opened 2026-04-18
Add Payload CMS sync runtime helper

mihaipxm:
## Summary
- extract Payload CMS client/logger/mapper/event-name assembly into a shared sync runtime helper
- refactor the Payload CMS plugin to build subscribers from that runtime instead of inlining setup
- add focused unit coverage for the new helper while keeping plugin behavior the same

## Testing
- git diff --check
- pnpm -C packages/plugins/payload-cms lint
- pnpm -C packages/plugins/payload-cms typecheck
- pnpm -C packages/plugins/payload-cms test


mihaipxm:

PR #173, opened 2026-04-18
Use shared route parsers in Netopia plugin routes

mihaipxm:
## Summary
- switch the Netopia plugin Hono routes to the shared Hono body parser
- keep the plugin aligned with the framework transport conventions without changing runtime behavior
- cover payment session start, collect, and callback payload parsing through the shared helper

## Testing
- git diff --check
- pnpm -C packages/plugins/netopia lint
- pnpm -C packages/plugins/netopia typecheck
- pnpm -C packages/plugins/netopia test


mihaipxm:

PR #184, opened 2026-04-18
Use shared optional body parsing in product version route

mihaipxm:
Summary:
- move the product version snapshot route onto the shared optional Hono body parser
- keep the existing route behavior where an empty body is allowed and normalized through the schema
- finish the remaining optional-empty-body transport cleanup in packages/products/src/routes.ts

Testing:
- git diff --check
- pnpm -C packages/products lint
- pnpm -C packages/products typecheck
- pnpm -C packages/products test
- full repo pre-push suite


mihaipxm:

PR #350, opened 2026-04-28
fix(templates): depend on @voyantjs/cloud-sdk directly, drop voyant-cloud wrapper

mihaipxm:
## Summary

Templates (`dmc`, `operator`) now depend on `@voyantjs/cloud-sdk` directly instead of going through the `@voyantjs/voyant-cloud` workspace wrapper. The SDK is the source of truth; we don't need another abstraction layer on top.

## Why

`@voyantjs/voyant-cloud` is a ~30-line wrapper around `@voyantjs/cloud-sdk` that:
- Reads `VOYANT_CLOUD_API_KEY` / `VOYANT_CLOUD_API_URL` / `VOYANT_CLOUD_USER_AGENT` from `env`
- Calls `createVoyantCloudClient(...)` from the SDK
- Throws if the API key is missing

That's it. It doesn't justify a separate published package — and the package has been failing to publish on npm for several release cycles (`E404 Not Found - PUT`). Templates can't depend on a package that doesn't exist on npm; downstream consumers cloning the templates would hit unresolved deps.

## What changed

**Per template** (`templates/dmc`, `templates/operator`):
- New `src/lib/cloud-client.ts` (~70 lines) — local `getVoyantCloudClient(env)` + `tryGetVoyantCloudClient(env)` + `VoyantCloudConfigError`. Identical behavior to the old wrapper.
- All template-side imports of `@voyantjs/voyant-cloud` updated to import from `./cloud-client.js`:
  - `src/lib/notifications.ts`
  - `src/lib/brochure-printer.ts`
  - `src/lib/video-uploads.ts`
  - `src/api/invitations.ts`
  - `src/api/auth/handler.ts`
- `package.json`: drop `@voyantjs/voyant-cloud`, add `@voyantjs/cloud-sdk@^0.4.0`

## What's left

The `@voyantjs/voyant-cloud` workspace package stays for non-template consumers (`@voyantjs/notifications`, `@voyantjs/plugin-netopia`, etc.). Whether they should also bypass it is a follow-up question — this PR is scoped to templates so the publish failure stops blocking template installs.

If we decide later that none of the wrappers (`voyant-cloud`, `vault`, `verify`) are pulling weight, we can privatize them in `.changeset/config.json` `ignore` and rewire all in-repo consumers similarly.

## Test plan

- [ ] CI typecheck stays green (verified locally: 177/177)
- [ ] Run `pnpm -F dmc dev` and `pnpm -F operator dev` — confirm Cloud SDK resolves and notifications/PDF/video uploads still work
- [ ] On next release, confirm `voyant-cloud@*.*.*` either publishes (if the npm-side issue gets resolved) or fails without blocking the rest of the release train


mihaipxm:
Superseded by upcoming PR — see explanation below.

This PR was opened 2026-04-28 before `@voyantjs/cloud-sdk` v0.6.0 shipped (2026-04-29 11:35) with native `getVoyantCloudClient`/`tryGetVoyantCloudClient`/`VoyantCloudConfigError`/`VoyantCloudEnv` (per voyantjs/cloud-sdk#5). At the time, the only path was to copy the env-helper into each template as `src/lib/cloud-client.ts` — that's no longer necessary.

The follow-up PR uses the SDK helpers directly (no local copies), and also covers `apps/dev/`, `packages/notifications/`, and deletes `packages/voyant-cloud/` itself — which this PR didn't touch. Net: ~150 fewer LOC of duplicated cloud-client wiring, and the wrapper package is fully retired.

PR #445, opened 2026-05-07
Split CI verify into parallel checks and build jobs

mihaipxm:
## Summary

Restructure the `verify` job in `.github/workflows/ci.yml` into two jobs that run concurrently:

* **checks**: lint, typecheck, test, architecture, i18n
* **build**: turbo build, package-export checks, publish-tarball checks

Both share setup-node + pnpm install. Turbo + pnpm remote caches make the duplicated install/build close to free on warm runs. Wall time becomes `max(checks, build)` instead of the previous `sum(checks, build)`.

## Heads up — branch protection

Splitting `verify` renames the required check. If branch protection currently requires `CI / verify`, update it to require both:

* `CI / checks`
* `CI / build`

Otherwise this PR will read as "missing required status" once merged.

## Scope changes since first push

The earlier version of this branch also bundled Phase 1 (REUSE_DIST gate) and Phase 2 (skip prepack tsc + bulk-extract) for `verify:publish-tarballs`. Those landed separately in #439 and are already on main. A `turbo.json` typecheck-input override was also dropped after Codex flagged that excluding `package.json` would silently mask changes to `scripts.typecheck`, `type`, `exports`, etc. — turbo can't hash JSON sub-fields, so there's no surgical way to exclude only `version`. Accepting the typecheck cascade on `chore: release packages` runs is the correct trade-off.

## Test plan

- [ ] CI on this PR shows two parallel jobs (`checks` + `build`); both go green
- [ ] Total wall time is `max(checks, build)`, not the prior sum
- [ ] After merge, branch protection updated to require both new checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,13 +17,21 @@ concurrency:
 permissions:
   contents: read
 
+# Two jobs run in parallel. `checks` covers everything that only needs source
+# (lint/typecheck/test/architecture/i18n); `build` covers everything that
+# needs fresh dist (turbo build → package exports → publish tarballs). They
+# share the same setup; both rely on the pnpm + turbo remote caches to make
+# the duplicated install/build effectively free on warm caches.
+#
+# NOTE: branch protection rules need to require both `CI / checks` and
+# `CI / build` (the previous single `verify` job no longer exists).
+
 jobs:
-  verify:
+  checks:
+    name: checks
     runs-on: ubuntu-latest
     env:
       HUSKY: 0
-      # Turbo remote cache. Configure in GitHub Settings -> Secrets and variables
-      # -> Actions; without them turbo falls back to the local runner cache.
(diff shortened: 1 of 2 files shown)


mihaipxm:
Updated this PR after the Codex review.

Force-pushed to drop both the duplicated Phase 1+2 commits (those landed via #439) and the `turbo.json` typecheck-input override. Codex was right that excluding `package.json` entirely was unsafe: changes to `scripts.typecheck`, `type`, `exports`, `dependencies`, etc. would silently replay an old cache result. Turbo can't hash sub-fields of JSON, so there's no clean way to exclude only `version` without losing the others. Accepting the typecheck cascade on `chore: release packages` runs is the correct trade-off — release PRs are infrequent and other PRs are unaffected.

What's left in this PR is just the parallel-job split (Option C from the plan): `verify` → `checks` + `build` running concurrently. Branch protection still needs the heads-up — `verify` is gone, requires `CI / checks` and `CI / build` instead.

Issue #518, opened 2026-05-09
Wire TriggerOptions enforcement: debounce / delay / ttl / priority / concurrencyKey


mihaipxm:
## Background

`TriggerOptions` (`packages/workflows/src/trigger.ts:29-40`) declares 9 fields. As of today, **6 are declared-only** (no orchestrator code reads them) and 3 work:

| Field | Status | Evidence |
|---|---|---|
| \`idempotencyKey\` | declared-only — being wired in #514's PR1 | \`packages/workflows-orchestrator/src/orchestrator.ts:22-52\` ignores it |
| \`debounce\` | declared-only | no handler |
| \`delay\` | declared-only | no scheduling mechanism |
| \`ttl\` | declared-only | no expiration |
| \`priority\` | declared-only | no scheduler consumes it |
| \`concurrencyKey\` | declared-only | no enforcement |
| \`tags\` | wired | \`orchestrator.ts:29\` passes through |
| \`lockToVersion\` | wired | passes through to RunRecord |
| \`environment\` | wired | passes through |

This issue tracks the 5 still-unwired fields after #514's PR1 lands `idempotencyKey`. They're all the same shape (declared in SDK, ignored in orchestrator) but they're independent features and can be tackled in separate PRs.

### \`delay: Duration | Date\`
Trigger creates the run record but the orchestrator parks it on a synthetic DATETIME waitpoint with \`wakeAt = now + delay\` (or \`wakeAt = delay\` if Date). Existing time wheel handles wake. Run starts at \`wakeAt\`.

### \`ttl: Duration\`
Set \`workflow_runs.ttl_at = now + ttl\` (or DO-storage equivalent). A new sweeper (or extended time wheel) marks runs \`expired\` if they haven't started by \`ttl_at\`. Runs already in-flight are unaffected.

### \`debounce: { key: string; delay: Duration; mode?: \"leading\" | \"trailing\" }\`
A debounce table keyed by \`(workflowId, debounceKey)\` holds the most recent trigger. Subsequent triggers within \`delay\` either update the held trigger (\`trailing\`, default) or are dropped (\`leading\`). Time wheel fires the held trigger at \`delay\` after the most recent update.

### \`priority: number\`
Adds a \`priority\` column on the run record; the time wheel's claim query orders by \`priority DESC, due_at ASC\`. Higher priority runs are claimed first when contention exists.

### \`concurrencyKey: string\`
[...]

Issue #517, opened 2026-05-09
Implement schedule firing for ScheduleDeclaration (cron / every / at)


mihaipxm:
## Background

`@voyantjs/workflows` declares `WorkflowConfig.schedule` (`packages/workflows/src/workflow.ts:47-58`):

```ts
export type ScheduleDeclaration = (
  | { cron: string }
  | { every: Duration }
  | { at: string | Date }
) & {
  timezone?: string
  input?: unknown | (() => unknown | Promise<unknown>)
  enabled?: boolean
  overlap?: "skip" | "queue" | "allow"
  environments?: ("production" | "preview" | "development")[]
  name?: string
}
```

A workflow can be declared as `schedule: [{ cron: "0 * * * *" }]` (or an array of mixed shapes), but **nothing fires anything on that cadence.** No cron evaluator exists in `packages/workflows-orchestrator/src/` or in `packages/workflows-orchestrator-cloudflare/src/`. `apps/workflows-local-dashboard/src/lib/api.ts` polls a hypothetical `/api/schedules` endpoint that has no handler.

Result: scheduled workflows silently never run.

## Scope

A scheduler that:

1. Reads schedule declarations from the manifest (collected by `createApp()` per the `feat/workflows-runtime` design).
2. Computes each schedule's next-fire time, filters by `environments[]`, respects `enabled`.
3. Fires `driver.trigger(workflow, input, opts)` at each cadence point.
4. Applies `overlap` semantics to in-flight runs of the same workflow.
5. Resolves `input` (literal or async function — function form runs at fire time).

## Implementation surface

Two viable approaches:

- **Mode 2 (Postgres)**: a `workflow_schedules` table with `next_fire_at` indexed; the existing time wheel from §7.2 of the design doc claims rows via SKIP LOCKED, fires triggers, computes the next slot, updates. Same primitive that wakes parked DATETIME waitpoints.
- **Mode 1 (CF DO)**: a singleton scheduler DO with alarms; on each alarm fire, evaluate due schedules from the manifest, trigger runs, schedule the next alarm. Or: a small Cloudflare Cron Trigger pointing at a route that does the same evaluation.

Cron parsing: pick a small dependency-free library (or hand-roll for the ~5 standard shapes Voyant needs). `every: Duration` is simpler — fixed-interval. `at: Date` is one-shot.

## Tests

- Cron expression evaluation (truth table over common shapes).
- Schedule firing across all stores via compliance suite (parameterized).
[...]

Issue #516, opened 2026-05-09
Wire ConcurrencyPolicy enforcement in workflow orchestrator


mihaipxm:
## Background

`@voyantjs/workflows` declares `WorkflowConfig.concurrency` (`packages/workflows/src/workflow.ts:60-64`):

```ts
export interface ConcurrencyPolicy<TInput> {
  key?: string | ((input: TInput) => string)
  limit?: number
  strategy?: "queue" | "cancel-in-progress" | "cancel-newest" | "round-robin"
}
```

The orchestrator core does **not enforce any of it.** No code in `packages/workflows-orchestrator/src/` reads `ConcurrencyPolicy` or computes a concurrency key. Workflows that declare e.g. `concurrency: { key: (input) => input.tenantId, limit: 1, strategy: "queue" }` silently run unconstrained.

This is a silent footgun: the type surface invites operators to express constraints that have zero runtime effect.

## Scope

Wire end-to-end enforcement of the four declared strategies:

- **`queue`** — secondary triggers under the same key park until limit-1 finish; FIFO drain.
- **`cancel-in-progress`** — new trigger cancels the in-flight run under the same key, then starts.
- **`cancel-newest`** — new trigger is rejected/cancelled if a run already holds the key; existing run continues.
- **`round-robin`** — across multiple keys, distribute across the limit. (Lower priority; could defer if mechanism diverges from the others.)

`limit > 1` semantics: the first `limit` triggers run concurrently; further triggers respect the strategy.

## Implementation surface

Most of the work is in two places:

1. **Concurrency-key resolution at trigger time** — the orchestrator's `trigger()` (`packages/workflows-orchestrator/src/orchestrator.ts:61`) resolves `key` from input + workflow config, looks up active runs holding that key.
2. **Per-store enforcement adapters**:
   - **InMemory**: in-process Map keyed by concurrency-key → array of active runIds.
   - **Mode 2 (Postgres)**: a separate \`workflow_concurrency_holders\` table or a new \`concurrency_key text\` column on \`workflow_runs\` with index, queried under transaction at trigger time.
   - **Mode 1 (CF DO)**: a singleton 'concurrency map' DO per environment, or deterministic DO addressing by key. Choose at design time.

Cancellation strategies reuse existing \`cancel\` (\`orchestrator.ts:206\`) + compensation cascade.

## Tests

- Compliance suite gains \`'concurrency: queue strategy parks secondary triggers'\`, etc., parameterized over all stores.
[...]

Issue #514, opened 2026-05-09
Implement trigger.on() runtime in workflow orchestrator


mihaipxm:
## Background

`@voyantjs/workflows` exposes `trigger.on(eventName, { target, match, input })` for declarative event-driven workflow triggering. The declaration is collected into `WorkflowManifest.eventFilters` (per `packages/workflows/src/protocol/index.ts:85`).

**No runtime consumes it.** Greppable evidence:
- `EventFilter` and `trigger.on` appear only in the SDK declaration (`packages/workflows/src/trigger.ts`) and the protocol shape (`packages/workflows/src/protocol/index.ts`).
- Zero references in `packages/workflows-orchestrator{,-cloudflare,-node}` or `apps/workflows-{tenant,orchestrator,selfhost-cloudflare}-worker`.
- `WorkflowsClient` proxy at `packages/workflows/src/trigger.ts:102-111` throws on every method call: *"requires the Voyant Cloud client. Install + configure it via @voyantjs/client"* — that client doesn't exist in this monorepo.

Result: `trigger.on(...)` calls compile and register a manifest entry but never fire a workflow at runtime. This is a significant capability gap — every event→workflow flow we'd want (snapshot capture, pricing reindex, channel push, promotions reindex) has to fall back to either polling (DB flag + scheduled poll) or inline work in the operator's event subscribers (unsafe on Workers for any unbounded job).

## What the runtime needs

1. **Event ingestion endpoint** on the orchestrator (e.g., `POST /api/events { type, payload, scope?, metadata? }`)
2. **Filter matcher** — walk `WorkflowManifest.eventFilters`, find entries with matching `eventType`, evaluate `matchExpression` if present, build the workflow input via the `input` mapper
3. **Workflow trigger** — for each match, call the existing `POST /api/runs` flow (or invoke directly within the orchestrator) with the computed input
4. **Operator template integration** — opt-in bridge subscriber that forwards `eventBus.emit(...)` to the orchestrator's event ingest, configured via `WORKFLOWS_ORCHESTRATOR_URL` (or whatever wiring shape the team prefers)
5. **Docs + one worked example** — pick a small consumer to migrate (the easiest is probably promotions reindex once #2 lands)

## Open questions for design

- **Persistence**: synchronous fire-and-forget HTTP from the operator, or queued ingest with a persistent log? Persistent gives durability but adds storage; fire-and-forget is simpler and matches the current event bus's at-most-once semantics.
[...]

PR #523, opened 2026-05-09
feat: createApp() workflow wiring + Module/Plugin collection — PR4 of #514

### LINKED ISSUE #515: Move promotions affected="all" reindex from log+skip to a workflow
## Background

PR #513 (promotions module, #497) ships with one explicit limitation: when a promotion mutation or boundary-scheduler firing has `affected.kind === \"all\"` (a global / market / audience scope change), the catalog-bridge subscriber and the boundary-scheduler cron handler **log + skip** instead of reindexing.

The reason is correctness, not laziness:
- Inline enumeration of every owned product in a Cloudflare Workers subscriber is unsafe — CPU and wall-time limits make any unbounded loop a timeout risk for non-trivial catalogs.
- The architecture doc (`docs/architecture/promotions-architecture.md` §15.1 + §14) acknowledged this as a v1 limitation.

Today, operators run `pnpm exec tsx scripts/reindex.ts products` manually after creating / expiring a global / market / audience-scoped offer. That works but isn't \"end-to-end\" — the user-facing promise of \"badges + strikethrough prices appear automatically\" doesn't hold for these scope kinds.

## What this issue does

Move the bulk reindex into a `@voyantjs/workflows` workflow with `defaultRuntime: \"node\"` so the work runs in the tenant worker (Node / container — no Workers limits), triggered automatically when an `affected=\"all\"` `promotion.changed` event fires.

## Concrete plan

1. Define the workflow in `templates/operator/src/workflows.ts`:

   \`\`\`ts
   const promotionsReindexAll = workflow<{ offerId: string; source: string }, { reindexed: number }>({
     id: \"promotions.reindex-all-products\",
     defaultRuntime: \"node\",
     async run(input, ctx) {
       const db = getDb()
       // Same enumeration scripts/reindex.ts uses — SELECT id FROM products,
       // build indexer + builder, reindexEntity per product. Returns count.
     },
   })
   \`\`\`

2. Add the declarative trigger:

   \`\`\`ts
   trigger.on(\"promotion.changed\", {
     target: promotionsReindexAll,
     match: (p) => p.affected.kind === \"all\",
     input: (p) => ({ offerId: p.offerId, source: p.source }),
   })
   \`\`\`

3. Delete the log+skip branches in:
   - `templates/operator/src/api/catalog-bridge.ts` (the `promotion.changed` subscriber)
   - `templates/operator/src/api/promotion-scheduled.ts` (the cron handler)

4. Update the architecture doc § Known limitations to mark this as resolved.

## Blocked by

- voyantjs/voyant#514 (`trigger.on()` runtime). Until that lands, the workflow definition + `trigger.on(...)` declaration would compile + register the manifest entry but never fire. We can ship the workflow definition early and just leave the log+skip in place until the runtime is wired — but the cleanup step (deleting log+skip) needs #514 done first.

## Alternative considered: polling workflow + DB flag

Drop a small `promotional_offer_reindex_pending` flag table. Subscribers INSERT a flag; a scheduled workflow drains every 1 min. Self-contained, no orchestrator-runtime dependency. Decided against because:
- Adds a new table + schema migration just for this
- Adds 1-min latency vs near-immediate event triggering
- The `trigger.on()` runtime (#514) is foundational and worth doing properly — polling here would be throwaway code in a few weeks

## Acceptance

- Operator creates a global 20% offer → storefront cards refresh within ~10s (workflow trigger + reindex + Typesense propagation)
- Operator archives a global offer → cards lose the badge similarly fast
- Manual `pnpm exec tsx scripts/reindex.ts products` no longer required for global / market / audience scope changes

mihaipxm:
## Summary

PR4 of #514. Closes the framework-wiring piece. `createApp()` now collects `workflows` + `eventFilters` from modules + plugins, invokes the configured `DriverFactory` with framework deps (`services` + `logger`), and inside the existing lazy bootstrap path registers the manifest and installs an EventBus forwarder that routes emitted events into `driver.ingestEvent`. Adds `app.ready()` so headless / sibling-process deployments can fire the bootstrap before any HTTP traffic.

**Stacked on #522 (PR3).** Base: `feat/workflows-runtime-pr3`.

### `VoyantAppConfig` adds `workflows`

```ts
createApp({
  modules: [promotionsModule, ...],
  workflows: {
    driver: createNodeStandaloneDriver({ db }),  // DriverFactory
    environment: "production",                    // or preview / development
    projectId: "operator",                        // single-tenant default: "default"
  },
})
```

`workflows.driver` is structurally typed (`(deps) => WorkflowDriver`) so the SDK package stays leaf — concrete factories from `@voyantjs/workflows-orchestrator{,-node,-cloudflare}` satisfy via TypeScript structural compat.

### `createApp()` rewrite (synchronous setup)

- Collects `collectedWorkflows` + `collectedFilters` from every `HonoModule.module.workflows` / `eventFilters` and from every `HonoBundle.workflows` / `eventFilters` (plugins).
- **Validates duplicate workflow ids across the app** — throws at construction time with a helpful message.
- When `workflows.driver` is set, invokes the factory now with `{ services, logger }` so the driver handle is available throughout construction. HTTP route handlers can resolve it via `c.var.workflowDriver`.

### Lazy bootstrap (existing `bootstrapPromise` path)

After the existing isolated module / plugin / extension bootstraps, calls `wireWorkflowRuntime`:
1. Builds a deterministic `WorkflowManifest` via `@voyantjs/workflows/events` `buildManifest()`.
2. Calls `driver.registerManifest({ environment, manifest })`.
3. Subscribes one EventBus forwarder per unique manifest `eventType`. Forwarder stamps `metadata.eventId` with a content-derived id when missing — so the framework's idempotency derivation gives stable run-dedup across retries.

**Fail-closed** per architecture doc §21.22 — `registerManifest` failures reject the bootstrap promise and surface as 503 on the next request. Module bootstraps stay isolated (existing behavior); workflow registration does not.

### `app.ready()`

Returns the bootstrap promise (creating it eagerly with empty bindings if not yet fired). Tests + Mode 2 sibling-process deployments call this so the time wheel and manifest registration kick off without traffic. Production code never has to.

### `HonoBundle` / `Plugin` shape

Both gain `workflows?` + `eventFilters?` mirroring `Module`'s. `expandHonoBundles` already collects modules / extensions / subscribers / links; `createApp` pulls workflows + eventFilters off the original `HonoBundle` list directly.

### Dependencies

`@voyantjs/hono` adds `@voyantjs/workflows` (for `buildManifest` + `EventFilterRuntimeEntry` types) and `@voyantjs/workflows-orchestrator` (devDep for the test harness).

## What's deferred

PR4 originally scoped a promotions migration that closes #515 (delete `affected.kind === "all"` warn blocks in `templates/operator/src/api/{catalog-bridge,promotion-scheduled}.ts`; ship `bulkReindexProducts` workflow in `@voyantjs/promotions`). The promotions module + operator template work both live on the `feat/promotions` branch (PR #513), not on `main`. So a small follow-up PR will close #515 once both this branch and #513 merge to `main`. Tracked as an internal todo; not required for #514's design surface.

## Test results

| Package | Tests | Notes |
|---|---|---|
| `@voyantjs/hono` | **58/58 ✅** | 52 prior + 6 new `app-workflows` integration tests |
| `@voyantjs/workflows` | 254/254 ✅ | No regression |
| `@voyantjs/workflows-orchestrator` | 76/76 ✅ | No regression |
| `@voyantjs/core` | 129/129 ✅ | No regression |

The 6 new integration tests cover:
1. Module ships workflow + filter; emit `eventBus` event with matching `data.kind=all` → run created via the forwarder.
2. Non-matching `where` → no run.
3. Unrelated event name → no run (no subscriber installed).
4. Duplicate workflow ids across modules → construction-time throw.
5. `app.ready()` is idempotent.
6. Apps without `workflows` config are unaffected (no driver, no forwarder, regular subscribers still fire).

`pnpm typecheck` clean across all touched packages.

## What this means for #514

With PR4 merged on top of PR1 → PR2 → PR3:

- `trigger.on()` works end-to-end. Modules + plugins ship workflows + filters; `createApp` collects them; events flow through the forwarder into the driver; runs trigger.
- All three deployment shapes (InMemory, Mode 2 / Postgres, Mode 1 / CF edge) pass the parameterized compliance suite.
- The optional HTTP ingest adapter is available for self-host Mode 2 + voyant-cloud's HTTP boundary.
- Audit-found engine gaps (#516 concurrency, #517 schedules, #518 TriggerOptions) are tracked separately and don't block #514.

The only remaining loose thread is the promotions migration that closes #515 — that follow-up depends on #513 landing, and is structurally trivial once it does (one workflow definition + one trigger.on declaration + delete two warn blocks).

## Test plan

- [ ] `pnpm -F @voyantjs/hono test` — 58 green.
- [ ] `pnpm -F @voyantjs/hono typecheck` — clean.
- [ ] Read `app-workflows.test.ts` standalone — best place to see the wiring end-to-end.
- [ ] Read the new lazy-bootstrap section in `app.ts` — about 50 lines of net additions inside the existing `ensureRuntimeBootstrapped`.
- [ ] Verify cross-package smoke: `pnpm -F @voyantjs/workflows test`, `pnpm -F @voyantjs/workflows-orchestrator test`, `pnpm -F @voyantjs/core test` — all green.

## References

- Issue: voyantjs/voyant#514.
- Design: voyantjs/voyant#519.
- Stack: voyantjs/voyant#522 (PR3 — Mode 1 + HTTP ingest) → voyantjs/voyant#521 (PR2 — trigger.on + DSLs) → voyantjs/voyant#520 (PR1 — driver foundations + Mode 2).
- Architecture: §10 (module + plugin declaration), §17 (auth/telemetry), §18 (createApp boot flow), §18.1 (sibling-process deployment), §21.19 (structural descriptors for cycle avoidance), §21.22 (fail-closed manifest registration).


mihaipxm:
## Review fixes pushed (commit `3d490305e`)

All 5 review findings addressed in a single fix-up commit on PR4. Changes touch code from PR1-PR4 — the stack ends up correct on `main` once all 4 PRs land.

| Finding | Fix |
|---|---|
| **P1.1** Postgres idempotency race | `RunRecordStore.tryInsert(record): { record, created }` — atomic across all 4 store impls. Postgres uses `INSERT … ON CONFLICT DO NOTHING RETURNING` + fallback SELECT. New compliance test fires 8 parallel triggers with same `idempotencyKey`, asserts body executes ≤ 1 time. **Verified against real Postgres + InMemory + CF in-process fake (with per-id queue mirroring CF DO single-threading).** |
| **P1.2** Mode 2 sleep not actually waking | `createNodeStandaloneDriver` now wires `createPostgresWakeupStore` + `createPersistentWakeupManager`, auto-starts the poller, calls `syncWakeupFromRecord` after every trigger / ingestEvent forward, stops the poller on `shutdown()`. New `sleep-resume.integration.test.ts` triggers a workflow with `ctx.sleep("100ms")`, verifies it parks (status=waiting + `voyant_wakeups` row), waits for the poller to wake it, asserts `ctx.step` post-sleep ran exactly once. New options: `wakeupPollIntervalMs`, `wakeupLeaseMs`, `wakeupLeaseOwner`, `disableTimeWheel`. |
| **P2.1** Driver construction too early for env bindings | `VoyantWorkflowsConfig.driver` is now always `(bindings) => DriverFactory`. Mode 2 / InMemory wraps with `() => createXxxDriver({...})`; CF-edge uses `(env) => createCloudflareEdgeDriver({ env.* })`. Driver construction moves into the lazy bootstrap path with real `c.env`. New test verifies the deferral. |
| **P2.2** Random fallback eventId loses dedup | New `deriveStableEventId(envelope)` in `@voyantjs/workflows/events`: `evt_<safeName>_<safeAt>_<sha256(canonical(data)).slice(0,12)>`. Replaces random fallback in InMemory + Mode 2 + Mode 1 drivers + CF event-handler. New compliance test fires two `ingestEvent`s with byte-equal envelopes (no `metadata.eventId`), asserts same `eventId` + same `runId`. |
| **P2.3** Forwarder installed after module bootstraps | `wireWorkflowRuntime` moves to BEFORE the module/plugin/extension bootstrap loop. New test asserts a `module.bootstrap` that emits an event triggers the workflow run. |

## Test results

| Package | Tests | Notes |
|---|---|---|
| `@voyantjs/workflows-orchestrator` | 78/78 | +1 race + +1 stable-eventId compliance |
| `@voyantjs/workflows-orchestrator-cloudflare` | 67/71 | 4 architectural skips; CF in-process fake gained per-id queue |
| `@voyantjs/workflows-orchestrator-node` | 39 unit + **25 against real Postgres** (was 21) — 23 compliance + 2 sleep-resume integration |
| `@voyantjs/hono` | 8/8 (was 6; +2 deferred-driver + bootstrap-emit) |
| `@voyantjs/workflows` | 254/254 |
| `@voyantjs/core` | 129/129 |

`pnpm typecheck` clean across all touched packages. Pre-existing `selfhost-server-postgres` entry-loader failure is unchanged on this branch (verified against `origin/main` in the original PR1 work).

## Architectural notes worth flagging

- **Single function-of-bindings shape for `driver`** (P2.1) is a small but visible API change for the Mode 2 case: existing test fixtures wrap with `() => factory` instead of using the factory directly. The unambiguous shape avoids the heuristic-based detection that an earlier draft tried.
- **CF in-process fake with per-id queue** (P1.1 test fixture): mirrors production CF DO single-threading guarantee. Without it, the fake races concurrent fetches even though real CF wouldn't. Necessary for the race compliance test on Mode 1.
- **Mode 2 vitest config `fileParallelism: false`**: integration test files share `TEST_DATABASE_URL`; parallel `TRUNCATE`s in `beforeEach` hooks raced sleep-resume's polling loop. Serial file execution keeps tests deterministic.

Stacked PRs (#520 → #521 → #522 → #523) should be reviewed/merged together — each individual PR diff still has the bugs the fix-up addresses, only the squashed end-state on `main` is correct.

Issue #1411, opened 2026-05-30
Define Admin API SDK contract for mobile and AI-native admin clients


mihaipxm:
## Summary

Define a stable Admin API contract and TypeScript SDK surface so non-web clients can operate against Voyant-powered admin deployments without depending on web-admin components or app-specific route incidental details.

This is intended as follow-up work after #1400 establishes the reusable contract-package pattern. The same contract discipline should then be applied to admin operations: bookings, CRM, finance, legal, products, workflows, settings, and cross-domain actions.

## Motivation

Voyant deployments already expose rich admin functionality through the web admin. A React Native + Expo mobile admin app, AI agents, automation workers, and other non-web clients need a stable programmatic surface that represents what an operator can do in a Voyant-powered admin panel.

The mobile app should not be only an AI chat app. It should let an authenticated operator perform native admin workflows directly, while Max/AI uses the same underlying actions and permissions.

To make that sustainable, mobile UI, web UI, Max tools, and platform brokers should not each invent their own per-route integration layer. They need a shared Admin API contract and SDK.

## Proposed Shape

Introduce an Admin API product surface with three layers:

1. **Admin API contracts**
   - Stable request/response schemas for admin operations.
   - Versioned DTOs independent of web components.
   - Shared error shapes, pagination, filtering, sorting, and mutation result semantics.
   - Permission and capability metadata per operation.
   - Action classification metadata, e.g. read, routine write, destructive, requires confirmation.

2. **Admin client SDK**
   - Framework-neutral TypeScript client usable from React web, Expo, Node, Cloudflare Workers, and Max tools.
   - Handles base URL, auth headers, retries, pagination, typed errors, and capability discovery.
   - Does not depend on React or web UI packages.

3. **Client adapters**
   - React Query hooks for web admin.
   - Expo/mobile-friendly hooks and helpers.
   - Max tool wrappers.
   - Voyant Cloud broker/server callers.

Package shape can be decided during design, but examples could be:

- `@voyantjs/admin-contract`
- `@voyantjs/admin-client`
- `@voyantjs/admin-react`
- `@voyantjs/admin-expo`

Alternatively, keep the current domain-family style and add `*-contract` / `*-client` packages where the surface is large enough.

## Candidate Operation Families

The first contract pass should prioritize high-value mobile and AI workflows:

- bookings: list, search, get, create, confirm, cancel, attach document, create checkout/payment link
- CRM: people/org search, get, create, update, merge-safe lookup helpers
[...]

PR #1446, opened 2026-05-31
Admin API contract + SDK (first slice) — design + bookings/finance

mihaipxm:
**Draft for design review.** This is the first slice of #1411 (the Admin API contract + SDK). The architecture is the substance here — please review **ADR-0003** before the code. The full operation catalogue, the server-side capabilities route, and an optional React Query adapter are explicit follow-ups (see the ADR roadmap; no separate Expo or Max adapter packages — both consume the client directly), not this PR.

## What's here

- **`docs/adr/0003-admin-api-contract-sdk.md`** — the design (the issue's first acceptance criterion). Grounds the decisions in how the admin surface actually works today (4-strategy auth, `resource:action` scopes, the `action-ledger` + approval/`202` confirmation flow, `{data,total,limit,offset}` pagination, `{error,code,...}` errors). Covers package boundaries, the operation-descriptor shape, action classification, capability discovery, alternatives, and the roadmap.

- **`@voyantjs/admin-contracts`** (zod-only) — `OperationDescriptor` + `defineOperation()`, action classification (`read | routine_write | destructive | requires_confirmation`), shared error/pagination envelopes, the capability-discovery descriptor, and the first operation catalogue:
  - bookings: `list`, `get`, `confirm`, `cancel`
  - finance: `invoices.list`, `invoices.get`, `payments.record`, `paymentLinks.create`

- **`@voyantjs/admin-client`** — framework-neutral `createAdminClient({ baseUrl, auth, fetch? })`. Runs in Expo/Node/Workers/Max tools (no React, no framework runtime). Handles auth (apiKey/bearer/custom), typed `AdminApiError`, query/body serialization, the `{data}`-vs-raw envelope, idempotency-key headers, and `client.capabilities()`.

## Key design decisions (call these out in review)

1. **Dedicated `admin-*` family** (`admin-contracts` / `admin-client` / later `admin-react` / `admin-expo`) over per-domain `bookings-client`/`finance-client`. Mobile + Max want **one** client with cross-module capability discovery; domains are namespaces inside it (`client.bookings.*`). Naming uses the `-contracts` suffix from ADR-0002.
2. **Operations are data** (descriptors), so the same definition drives the SDK call, a Max-tool wrapper (with risk gating from `classification`), the capabilities listing, and audit `actionKind` — no four parallel integration layers.
3. **Capability discovery** is defined in the contract now; the server `GET /v1/admin/_meta/capabilities` route is a follow-up.

## Verification

- `pnpm typecheck` — 134/134
- `pnpm test` — 135/135 (admin-contracts 6, admin-client 5, both with zod-validated I/O and a mock-fetch client suite)
- `pnpm lint` — 0 errors (1 pre-existing unrelated `products-ui` warning)

## Open question for reviewers

Package naming/shape is the one decision the issue explicitly left open. I went with the dedicated `admin-*` family (rationale in ADR-0003 §Alternatives). If you'd prefer per-domain client packages, it's a cheap change at this stage.

Closes none yet — this is the first slice toward #1411.

### DIFF
--- a/.changeset/admin-api-contract.md
+++ b/.changeset/admin-api-contract.md
@@ -1,0 +1,24 @@
+---
+"@voyantjs/admin-contracts": minor
+"@voyantjs/admin-client": minor
+---
+
+Introduce the Admin API contract + SDK (first slice).
+
+`@voyantjs/admin-contracts` defines admin operations as typed, versioned,
+transport-agnostic descriptors — `OperationDescriptor` + `defineOperation()`,
+action classification (`read | routine_write | destructive |
+requires_confirmation`), shared error/pagination envelopes, and a
+capability-discovery descriptor. It ships the first operation catalogue for
+bookings (list/get/confirm/cancel) and finance (invoice list/get, record
+payment, create payment link). Pure and zod-only.
+
+`@voyantjs/admin-client` is a framework-neutral client (`createAdminClient`)
+that executes those descriptors from Expo, Node, Workers, and Max/AI tools — no
+React or framework runtime deps. It handles auth (API key / bearer / custom),
+typed `AdminApiError`s, pagination, idempotency keys, and capability discovery.
+
+The architecture, package boundaries, and roadmap (server `_meta/capabilities`
+route, more domains, React/Expo adapters, Max-tool wrappers) are documented in
+`docs/adr/0003-admin-api-contract-sdk.md`. Web admin, mobile, Max tools, and
+brokers consume one surface, keeping permission and audit semantics consistent.
--- a/docs/adr/0003-admin-api-contract-sdk.md
+++ b/docs/adr/0003-admin-api-contract-sdk.md
@@ -84,12 +84,19 @@ identical across callers.**
   injection (API key **or** bearer/session), typed errors, pagination,
   idempotency keys, and capability discovery. **No React, no web-UI, no
   framework runtime deps** — runs in Expo, Node, Workers, and Max tools.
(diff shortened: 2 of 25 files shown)


mihaipxm:
Thanks — all three Codex findings were valid (real contract↔route drift) and are fixed in ff08bd911:

- **P1 (approval responses)** ✅ — `bookings.confirm`/`cancel` can return HTTP 202 with an approval-required envelope for agent/workflow callers. The client now parses that (`approvalRequiredSchema`) and throws a typed `AdminApprovalRequiredError` carrying `.approvalId`, instead of treating 202 as success and failing to parse the payload as a booking. Added a test for the 202 path.
- **P2 (invoiceType filter)** ✅ — dropped the unsupported `invoiceType` filter from `finance.invoices.list`; the input now mirrors `invoiceListQuerySchema` (`status`/`bookingId`/`personId`/`organizationId`/`currency`/`search`). (`invoiceType` remains an *output* field.)
- **P3 (payment-link amount fields)** ✅ — dropped `currency`/`amountCents` from `finance.paymentLinks.create` since the route derives both from the invoice balance; input now mirrors the payment-session provisioning schema (all optional).

These are exactly the descriptor↔route drift the ADR flags as a risk — I've noted the CI guard that would catch this class automatically (descriptor `path`/`method`/`input` asserted against the mounted route) as a roadmap item in ADR-0003 §Roadmap. Verification after the fixes: typecheck 134/134, admin tests 12/12, lint clean.

PR #1459, opened 2026-06-01
Admin capability-discovery route: GET /v1/admin/_meta/capabilities (#1411 5.1)

mihaipxm:
First **framework** slice of the Admin SDK (#1411). Makes `client.capabilities()` return live data so a client can introspect what a deployment supports.

## What it does

`createApp` serves a built-in `GET /v1/admin/_meta/capabilities` route (under the `/v1/admin/*` staff guard, like `/health`), returning:
- **enabled modules** (the mounted module names — `createApp` knows them)
- **operations** (the catalogue from `admin-contracts`)
- **contract/deployment version**
- the caller's resolved **actor + scopes**

## Layering

The route is mounted **only** when the deployment supplies the catalogue via the new `adminMeta` config:

```ts
createApp({
  adminMeta: { contractVersion: ADMIN_CONTRACT_VERSION, operations: operationCapabilities() },
})
```

`adminMeta` is typed **structurally**, so `@voyantjs/hono` stays fully decoupled from `@voyantjs/admin-contracts` (which transitively pulls bookings/finance-contracts — wrong direction for the foundational HTTP layer). The deployment injects it. Wired in `templates/dmc` as the reference.

## Verification

- `pnpm typecheck` — 144/144
- hono tests — 97 (incl. 2 new: route serves capabilities for staff + adminMeta; not mounted when `adminMeta` omitted)
- `pnpm verify:release-train` — 136 packages on 0.97.0
- lint clean

Also updates ADR-0003 (roadmap item 1 → done; clarifies the Max product is Voyant Cloud, not framework).

Note: the pre-commit agent-quality check flags pre-existing issues in `app.ts` (file-size, existing unsafe-casts/lint-disables) surfaced because the file changed — report-only, none added by this PR.

### DIFF
--- a/.changeset/admin-capabilities-route.md
+++ b/.changeset/admin-capabilities-route.md
@@ -1,0 +1,18 @@
+---
+"@voyantjs/hono": minor
+---
+
+Add the admin capability-discovery route — `GET /v1/admin/_meta/capabilities`.
+
+`createApp` now serves a built-in capabilities route (under the `/v1/admin/*`
+staff guard) when the deployment supplies the operation catalogue via the new
+`adminMeta` config: `{ contractVersion, deploymentVersion?, operations }`. It
+returns the enabled modules, the operation catalogue, the contract/deployment
+version, and the caller's resolved actor + scopes — so the admin SDK's
+`client.capabilities()` returns live data.
+
+`adminMeta` is typed structurally, so `@voyantjs/hono` stays decoupled from
+`@voyantjs/admin-contracts`; deployments inject the catalogue from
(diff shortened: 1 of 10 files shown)


mihaipxm:
Fixed the P1 in 0cc1a2cbf: `/v1/admin/_meta/*` is now exempt from the API-key module-scope check in `requireActor` — any authenticated API key can read the discovery endpoints (which report the key's own granted scopes), while module scopes stay enforced on every real admin route. `_meta` is a reserved namespace (no module is named `_meta`). Added two require-actor tests: a `bookings:read` key reaches `/v1/admin/_meta/capabilities` (200) but still 403s on `/v1/admin/finance/*`.

PR #1460, opened 2026-06-01
Add admin-contracts descriptor consistency guard (#1411 slice 5.4)

mihaipxm:
A pure, dependency-free **descriptor consistency guard** for the Admin API contract (`@voyantjs/admin-contracts`). A vitest test walks every `OperationDescriptor` in `allOperations` and asserts it is well-formed and internally consistent:

- **Unique operation ids** across the registry.
- **Module segment matches id prefix** — e.g. `finance.invoices.list` → `/v1/admin/finance/...`.
- **Valid `method` + `classification`**.
- **Canonical `resource:action` scopes** — mirrors the API-key scope grammar (hyphens allowed, so `bookings-pii:read` is accepted).
- **`path()` reproduces `pathTemplate`** once each `:param` is substituted with a distinct sample value — catching omitted/reordered/unsubstituted params, not just a leftover `/:`.

This is slice **5.4** of #1411. #3 already eliminated the input-schema drift class (admin inputs derive from the module-contracts route schemas); this guard closes the remaining authoring-drift class at the layer where the descriptors live. The complementary live route-existence check is deferred to a deployment test.

Both Codex P2s addressed (scope grammar + path/template comparison).

> Note: the catalogue expansion (CRM/legal/products) and the `@voyantjs/admin-react` package (slices 5.2/5.3) are **not** in this PR — they land separately. An earlier edit briefly mis-scoped this title; corrected.

### DIFF
--- a/.changeset/admin-contracts-consistency-guard.md
+++ b/.changeset/admin-contracts-consistency-guard.md
@@ -1,0 +1,11 @@
+---
+"@voyantjs/admin-contracts": patch
+---
+
+Add a descriptor consistency guard (test): asserts every admin operation
+descriptor is well-formed and internally consistent — unique ids, an
+`/v1/admin/<domain>` path matching the operation's id prefix, a valid
+method/classification, `resource:action` scopes, and a `path()` builder that
+substitutes every template param. Catches the authoring-drift class that makes a
+descriptor diverge from the API surface. (The complementary live route-existence
+check belongs in a deployment test; #1411 5.4.)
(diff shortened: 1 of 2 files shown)


mihaipxm:
Both addressed in f067b32bf:

1. **Hyphenated scopes** — the scope assertion now mirrors the canonical API-key grammar (`permissionStringPattern` in `packages/types/src/api-keys.ts`): `^[a-z][a-z0-9-]*:[a-z][a-z0-9-]*$`. PII-aware scopes like `bookings-pii:read` now pass. (`*` wildcards stay disallowed on purpose — a descriptor naming a wildcard scope is a bug.)

2. **Built path vs template** — instead of only checking for a leftover `/:`, the guard now substitutes each `:param` in `pathTemplate` with a *distinct* sample value (`sample-<name>`) and asserts `path(params)` reproduces that string exactly. A `/:id/confirm` builder returning `.../confirm` (omitted param), or one that reorders params, now fails.

PR #1464, opened 2026-06-01
Extract products validation to @voyantjs/products-contracts; complete admin products writes (#1411)

mihaipxm:
Extracts the products validation into the pure `@voyantjs/products-contracts` package and uses it to complete the products admin SDK surface — the deferred follow-up noted in `products.ts`.

## Extraction (mirrors bookings/finance/crm/legal → `*-contracts`)

The products validation cluster moves out of the runtime `@voyantjs/products` package into the pure `@voyantjs/products-contracts`:

- **Moved:** `validation`, `validation-core`, `validation-public`, `validation-shared`, `validation-config`, `validation-content`, `validation-catalog`.
- **Made pure:** the cluster's *only* external imports were two `@voyantjs/db` helpers in `validation-shared.ts` (`booleanQueryParam`, `typeIdSchema`) — repointed to `@voyantjs/schema-kit`. Every product enum is already defined inline, so the package stays zero-runtime (zod + schema-kit only).
- **Zero breakage:** the moved files become one-line re-export stubs in `@voyantjs/products`, so every existing import path keeps working unchanged — `@voyantjs/products/validation`, `@voyantjs/products/public-validation`, and the internal `./validation-*` paths. Verified across the downstream consumers (catalog, sellability, storefront, promotions, products-ui, catalog-react, octo).
- products-contracts exports the seven validation subpaths and gains a `@voyantjs/schema-kit` dependency.

## Admin SDK — products writes (#1411)

With validation now in a pure package, `admin-contracts` can derive from it:

- **`products.create`** / **`products.update`** ← `insertProductSchema` / `updateProductSchema`
- **`products.delete`** (scope `products:delete`)
- **`products.list`** now derives from `productListQuerySchema` (was a hand-mirrored subset) — so the SDK advertises exactly the route's filters
- `admin-client` gains typed `products.create` / `update` / `delete` methods
- Removed the "writes deferred" note from `products.ts`

The generic `@voyantjs/admin-react` hooks pick all of this up automatically (descriptor-driven).

## Verification

- products-contracts 4 / admin-contracts 9 / admin-client 10 tests — green.
- Targeted typecheck across the extraction + downstream consumers, full-workspace `typecheck` (pre-commit) and `test` (pre-push), lint, `verify:release-train` (137 pkgs @ 0.99.0), and `verify:package-exports` all pass.


mihaipxm:
Addressed in d3aa1c5c6. Good catch — this is the mirror of the earlier filter-drift class.

`productsService.listProducts` implements every `productListQuerySchema` filter **except `taxClassId`** (it's accepted by the query schema but never turned into a predicate). So the derived list input now omits it:

```ts
export const productsListInputSchema = productListQuerySchema.omit({ taxClassId: true })
```

Still derived from the canonical schema (no field-name drift), but it no longer advertises a filter that silently returns unfiltered results. Added a regression assertion: `taxClassId` is dropped while the real filters (`productTypeId`, `supplierId`, …) are kept.

PR #2682, opened 2026-06-30
Commit typed currency in price catalog form (and fix label associations)

mihaipxm:
## What
Settings → Price Catalogs "Add Catalog" let you type a currency code (e.g. `EUR`) but persisted a blank currency with no error. The dialog used a local combobox whose committed value only changed when a row was picked from the list — typed text only filtered, it was never bound to the form value (issue #2650).

`currency_code` is intentionally nullable (NULL = multi-currency catalog), so forcing "required" would break a legitimate case. Per the issue's first option, the fix commits the typed value.

## Change
- `@voyant-travel/ui` `CurrencyCombobox`: commit a fully-typed ISO code (case-insensitive, must be a known 3-letter currency) even when the matching row is never selected, so typed input is never silently dropped. Additive new optional props `id` (for label association) and `emptyLabel`. Strict improvement for every currency picker.
- `@voyant-travel/commerce-react` price catalog form: reuse the shared `CurrencyCombobox` (drops a duplicate local one) and associate every field's `<Label htmlFor>`/`id` (addresses the issue's label-association note).

## Verification
- `@voyant-travel/ui` + `@voyant-travel/commerce-react`: typecheck / lint / test — pass (5 new combobox tests: typed commit, lowercase normalization, no-commit on partial/invalid, clear→null, id forwarding)
- `pnpm i18n:check` — pass (reused existing keys)

Fixes #2650

### DIFF
--- a/.changeset/price-catalog-currency-validation.md
+++ b/.changeset/price-catalog-currency-validation.md
@@ -1,0 +1,6 @@
+---
+"@voyant-travel/commerce-react": patch
+"@voyant-travel/ui": patch
+---
+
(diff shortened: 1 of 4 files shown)


mihaipxm:
@codex review

mihaipxm:
@codex review

PR #2687, opened 2026-07-01
Honor catalog:search scope on admin catalog search

mihaipxm:
## What
An API token with the `Catalog read` preset (which includes `catalog:search`) got `403` on `POST /v1/admin/catalog/search`, even though the same token could read products and sessions worked fine (issue #2649).

## Root cause
The API-key scope guard in `packages/hono/src/middleware/require-actor.ts` derives the required *action* purely from the HTTP method: `POST → write/trigger/relay`. Catalog search is exposed as `POST` (it takes a complex JSON body) but is a **read-family** operation, so the guard only ever accepted `catalog:write`-family scopes — a `catalog:search`/`catalog:read` token was rejected.

## Change
Make the action derivation path-aware: for `/search` path segments, additionally accept the `search` and `read` actions. Purely additive and scoped to `/search` routes — every non-search `POST` (product writes, bookings, suppliers, pricing) stays gated on `write` exactly as before. Applied at both guard call sites (api-key and staff-session RBAC).

## Verification
- `pnpm --filter @voyant-travel/hono` typecheck / lint / test — pass (251 tests). New regression tests: `catalog:search`/`catalog:read` (and bare `catalog:search`) authorized for admin+public catalog search; a search/read-only token still `403` on `POST /v1/admin/products` and `/v1/admin/bookings`.

Fixes #2649

### DIFF
--- a/.changeset/catalog-search-scope-api-key.md
+++ b/.changeset/catalog-search-scope-api-key.md
@@ -1,0 +1,5 @@
+---
+"@voyant-travel/hono": patch
+---
+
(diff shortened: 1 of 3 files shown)


mihaipxm:
@codex review

mihaipxm:
@codex review

PR #2688, opened 2026-07-01
Surface quote failures in storefront checkout; fix render-phase setState

mihaipxm:
## What
In the storefront card checkout (`BookingJourney`), when the connector adapter's quote request 500s, the shopper could still reach Review and clicking `Confirm booking` did nothing — no modal, no request, no error. There was also a `Cannot update a component (BookingJourney) while rendering a different component (PaymentStep)` warning (issue #2645).

## Root cause
- A failed quote leaves `quote.data` null, so `available = quote.data?.available !== false` read as `true` and the step gates stayed satisfied — Review was reachable. `onConfirm` then began with `if (!quote.data?.quoteId) return` — a bare early return → silent no-op.
- `PaymentStep` called the parent's `setDraft` during render to snap the payment intent → the render-phase update warning.

## Change
- Consume `quote.error` as a hard block: gate Next/Confirm (`canAdvance`/`canCommit`) while the quote is failing, and render a recoverable `role="alert"` banner with a **Retry pricing** button (re-runs the query) in both wizard and stacked layouts.
- Make `onConfirm` non-silent: with no valid quote it sets a visible `confirmError` (`validation.quoteUnavailable`) instead of returning; cleared once a fresh quote settles.
- Move `PaymentStep`'s intent-snap `setDraft` into a `useEffect`.
- New i18n keys (`quoteFailed`/`retryQuote`/`quoteUnavailable`) in en + ro.

## Verification
- `pnpm --filter @voyant-travel/bookings-react` typecheck / lint / test — pass (84 tests, incl. a new quote-failure regression test)
- `pnpm i18n:check` — pass

Fixes #2645

### DIFF
--- a/.changeset/storefront-quote-failure-handling.md
+++ b/.changeset/storefront-quote-failure-handling.md
@@ -1,0 +1,5 @@
+---
+"@voyant-travel/bookings-react": patch
+---
+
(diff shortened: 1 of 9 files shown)


mihaipxm:
@codex review

mihaipxm:
@codex review

mihaipxm:
@codex review

PR #2689, opened 2026-07-01
Record supplyModel derived-vs-first-class decision (ADR-0010)

mihaipxm:
## What
Issue #2644 asked to *evaluate* promoting `supplyModel` from derived to a first-class field, and explicitly to **record the decision** rather than leave it implicit in a code comment. No code promotion is required unless a vertical needs a supply model the `bookingMode` derivation can't express.

## Change (docs only)
- Add `docs/adr/0010-supply-model-derived-from-booking-mode.md` — records the decision to **keep `supplyModel` derived from `bookingMode`**, with context (post-#2329 enforcement), both options + trade-offs, rationale, the explicit revisit trigger, and the touch points to change if/when promoted.
- Point the `deriveProductSupplyModel` comment in `packages/inventory/src/service-catalog-plane.ts` at the ADR (comment-only; behavior unchanged).

No schema column, migration, or authoring field added — the decision is to keep it derived.

Fixes #2644

### DIFF
--- a/docs/adr/0010-supply-model-derived-from-booking-mode.md
+++ b/docs/adr/0010-supply-model-derived-from-booking-mode.md
@@ -1,0 +1,124 @@
+# ADR-0010: Keep `supplyModel` derived from `bookingMode`
+
+- **Status:** Accepted (2026-07-01)
+- **Relates to:** [#2644](https://github.com/voyant-travel/voyant/issues/2644),
+  [#2329](https://github.com/voyant-travel/voyant/issues/2329) (supply-model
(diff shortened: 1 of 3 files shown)


mihaipxm:
Addressed the P2 in dad32070: updated the active design doc `docs/architecture/catalog-supply-models.md` to reconcile with this ADR. §2 now states `supplyModel` is **derived from `bookingMode`** (not authored), §5/§6 point at the derivation + ADR-0010, and §7's open question reframes as *when to promote* (deferring to the ADR's trigger + touch points). The design doc and ADR are no longer contradictory — the ADR is the single source of truth and the doc defers to it.

@codex review

PR #2691, opened 2026-07-01
feat(commerce): anonymous public market discovery endpoint

mihaipxm:
## What

Adds a **read-only, anonymous public market discovery endpoint** so storefront clients can resolve the supported markets/locales/currencies without admin auth.

`GET /v1/public/markets` returns each `active` market with its active locales and currencies, using a **narrow customer projection**:

- `id` — the **catalog-search scope key** (thread into catalog search as the `market` parameter)
- `code`, `name`, `regionCode`, `countryCode`, `defaultLocale`, `defaultCurrency`
- `locales[]` → `{ languageTag, isDefault }`
- `currencies[]` → `{ currencyCode, isDefault }`

The catalog runtime indexes and searches Typesense slices keyed by `market.id` (`loadCatalogSlices` → `market: market.id`; catalog search passes `body.market` straight into the slice/collection name). So `id` — not the human-facing `code` — is the value a storefront must send as `market` to hit the right collection for a non-default market. `code`/`name` are for display.

## Security / surface discipline

No admin/finance-internal fields are exposed — the projection deliberately **omits** `status`, `timezone`, `taxContext`, `metadata`, timestamps, and every finance/pricing internal (FX rate sets, exchange rates, price catalogs, product/channel rules, and the per-currency `isSettlement`/`isReporting` flags). The market `id` is exposed because it is the public scope key the search API already consumes (not a secret); a test asserts the internal fields do not leak on the wire.

The route is declared `anonymous: true` on `marketsHonoModule`; the framework anonymous allow-list (ADR-0008) is updated and its governing test (`anonymous-surface.test.ts`) asserts `/v1/public/markets` is included. Response carries `Cache-Control: public, s-maxage=300, stale-while-revalidate=600`.

## Scope

This is the **backend discovery slice** of #2643. Storefront UI selectors (market/currency pickers) and threading the selected scope through the rest of the storefront are follow-up work — hence `Refs #2643` rather than `Fixes`, keeping the issue open.

## Verification

- `@voyant-travel/commerce` typecheck + tests green (incl. `routes-public.test.ts` — projection + anonymous route + no-leak assertions).
- `@voyant-travel/framework` typecheck + `anonymous-surface.test.ts` green.
- OpenAPI framework specs regenerated (`framework-openapi.json` + `framework-storefront.json`) so the drift check passes.
- Changesets: `@voyant-travel/commerce` patch, `@voyant-travel/openapi` patch.

Refs #2643


mihaipxm:
@codex review

mihaipxm:
Fixed in 0606f4e7c. You're right — the catalog runtime keys indexed/searchable slices by `market.id` (`loadCatalogSlices` → `market: market.id`; search passes `body.market` straight into the slice/collection name), so returning only `code` gave clients no key that matches a non-default market's collection.

The projection now exposes **`id`** as the catalog-search scope key (thread into search as the `market` parameter), keeping `code`/`name` for display. Every other admin/finance-internal field stays omitted (`status`, `timezone`, `taxContext`, `metadata`, FX, `isSettlement`/`isReporting`). Changing the slice key to `code` instead would force a Typesense reindex, so exposing the existing scope key is the minimal correct fix. Regenerated the framework OpenAPI specs and updated the tests/changeset.

@codex review

mihaipxm:
Fixed in 5357e8e2d. Verified — the operator starter ships its own composed OpenAPI artifacts with a separate drift gate (`starters/operator/src/api/openapi.test.ts`), and `framework-openapi.json` + `framework-storefront.json` were indeed missing `/v1/public/markets`. Ran `pnpm --filter operator generate:openapi` and committed both (the gate's 5 tests now pass; `framework-admin.json` unchanged since the route is public). This is the operator-level counterpart to the `packages/openapi/spec/*` regeneration in the earlier commit.

@codex review

PR #2695, opened 2026-07-01
fix(operator): gate storefront trip composer behind a session

mihaipxm:
## Root cause

`/shop/composer` drives the `/v1/public/trips/*` create→price→reserve→checkout surface. Those routes are deliberately **not** on the anonymous allow-list: `listTrips`/`getTrip` have no per-customer scoping, so mounting them anonymously would leak every tenant trip. As a result the first action ("New trip") did `POST /v1/public/trips` and returned a raw `401 {"error":"Unauthorized"}`, leaving the composer stuck in its disabled state.

## Fix (option a — gate the UI)

Gate the composer route behind a session (`authClient.useSession()`). Anonymous visitors now get a clear sign-in prompt (with a link back to browsing) instead of a broken disabled/401 state. Browsing and booking individual items stay fully open to everyone.

I chose gating over "allow anonymous trip drafts" (option b) because trip create/list/get currently have no per-customer scoping — making them anonymous would be a tenant-data-leak, not a small change. A real customer sign-in/account flow (so the composer becomes usable, not just gated) is the proper follow-up and is tracked separately in #2621.

## Scope
Operator-starter-only, UI-level change (`shop_.composer.tsx` + storefront i18n en/ro). No package or API changes; no changeset needed (operator is changeset-ignored).

Fixes #2642

### DIFF
--- a/starters/operator/src/lib/storefront-i18n.tsx
+++ b/starters/operator/src/lib/storefront-i18n.tsx
@@ -230,6 +230,13 @@ const storefrontMessagesEn = {
   bookingJourney: {
     marketingLabel: "Email me occasional updates about new tours and promotions.",
   },
+  composer: {
+    gateTitle: "Sign in to build a trip",
+    gateBody:
+      "The trip composer saves a working draft against your account, so you need to be signed in before you can start one. Browsing and booking individual items stays open to everyone.",
+    gateSignIn: "Sign in to continue",
+    gateBrowse: "Back to browsing",
+  },
 }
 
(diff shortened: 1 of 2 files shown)


mihaipxm:
Addressed the P2 in a330c05c: the gate's sign-in link now passes `next=/shop/composer`, so a shopper who signs in from the composer gate returns to the composer (the sign-in route already honors `next` for post-login redirect). Also rebased on latest main — the earlier CI `checks` failure was an openapi-artifact drift from this branch predating #2685's legal-policy regen, not from this change.

@codex review

PR #2697, opened 2026-07-01
feat(storefront): anonymous market/currency/locale scope selection

mihaipxm:
## What

Completes the storefront side of public market discovery (#2643). The backend `GET /v1/public/markets` endpoint shipped in #2691; this adds the customer-facing **selection** and threads the chosen scope through search + detail.

### `@voyant-travel/storefront-react` (packaged data layer)
- `useStorefrontMarkets` hook + `listStorefrontMarkets` operation + `getStorefrontMarketsQueryOptions`, fetching `GET /v1/public/markets` and validating against a local `storefrontMarketSchema` mirroring the public projection (`id`, `code`, `name`, `regionCode`, `countryCode`, `defaultLocale`, `defaultCurrency`, `locales`, `currencies`). 5-min `staleTime` matching the endpoint's `s-maxage=300`.

### operator storefront (app wiring)
- `StorefrontScopeProvider`: localStorage-persisted market/locale/currency scope, SSR-safe, resets locale/currency to the market defaults on market select.
- `StorefrontMarketSelector`: market / currency / language selects in the storefront chrome; picking a market constrains currency/language to what that market supports.
- Threads the selected scope into catalog search (`shop.tsx`: `market = scope.marketId`, `locale`) and the product-detail/booking-quote pages.

The market **`id`** is the catalog-search scope key (per #2691). Everything falls back to the runtime default scope when nothing is selected, so the default experience is unchanged.

## Verification
- `@voyant-travel/storefront-react` tests pass; operator tests pass (added `StorefrontScopeProvider` wrapping to the two isolated detail-page render tests that now read scope).
- Changeset: `@voyant-travel/storefront-react` patch.

Fixes #2643


mihaipxm:
Both findings fixed in 42f66790, and the branch is rebased on latest main (the earlier openapi drift was because this branch predated #2685's legal-policy artifact regen).

- **P1 (checkout scope):** `BookingJourney` now takes an optional `scope` prop and forwards it to its live quote; `StorefrontBookingJourney` reads the shopper's selected market/currency/locale from the `(storefront)` scope provider and passes it in. Checkout now prices in the same scope as browse/detail.
- **P2 (requote on scope change):** `useBookingQuote` now includes scope in its React Query key, so changing market/currency re-quotes instead of serving the stale price.

`scope` is optional and defaults to the prior per-surface behavior, so admin surfaces are unaffected. catalog-react + bookings-react test suites pass; changeset added for both.

@codex review

mihaipxm:
Fixed in 49636558a. The three detail pages now pass `locale`/`market`/`currency` from the selected scope as query params to the `/v1/public/.../content` endpoints (which honor those before falling back to `Accept-Language`) and include scope in the content query key, so a persisted or changed selection localizes the detail content — not just the quote. `fetchContent` gained an optional `scope` arg that appends the params. Applied to product, accommodation, and cruise detail pages.

@codex review

mihaipxm:
Fixed in 4c413471a. The storefront checkout handler now sends the live `context.quoteId` in the `/v1/public/catalog/book` request. The server prefers an explicit `quoteId` over resolving the draft's stored `currentQuoteId` (which the wizard only persists on step transitions), so a shopper who changes market/currency mid-journey books the quote they're actually looking at rather than a stale-scope one. Falls back to draft resolution when there's no live quote yet.

@codex review

mihaipxm:
Fixed in f5ded711. Made the quote placeholder scope-aware: `useBookingQuote` now keeps the previous quote across draft edits (no flashing) but drops to `null` across a scope change, so `quote.data?.quoteId` is null while the re-scoped quote is in flight and the confirm guard blocks booking until it lands. This closes the change-scope-then-immediately-confirm window. Admin surfaces never change scope, so their keep-previous behavior is unchanged; catalog-react tests still pass.

@codex review

mihaipxm:
Fixed in 7960f71d6. Good catch — this was a regression from the previous commit. `/book` now loads the draft whenever `draftId` is present (even alongside an explicit `quoteId`), so its payload still feeds `engineParametersFromDraft` and the selected departure/room/pax/traveler options are preserved. The explicit `quoteId` only overrides which quote is booked; it falls back to the draft's `current_quote_id` when no `quoteId` is sent, and to the `quoteId` when the draft is absent. Catalog booking-engine tests (410) still pass; changeset added for @voyant-travel/catalog.

@codex review

mihaipxm:
Valid, but this one needs backend support rather than a client param pass-through: the departure-slots reader `listAvailabilitySlots(db, productId, todayIso)` takes no market/locale/currency, and `/v1/public/catalog/slots` doesn't accept them — making departures vary by market is a server capability (reader signature + cross-package implementations + query schema), not a scope-threading tweak. The core selection path (discovery + selector + scoped search/detail/content/quote/checkout) is complete here, so I've filed #2701 to track market-scoped departure slots as a targeted follow-up.

PR #2710, opened 2026-07-01
fix(storefront): stop surfacing unrenderable cruises in customer shop

mihaipxm:
## Problem

Storefront customer search (`/shop?vertical=cruises`) surfaces cruise ids, but the cruise detail page fails content resolution:

- Owned cruise (`cru_…`) → public cruise content endpoint returns **404**.
- Demo/non-sourced cruise (`cdmi_…`) → **400** `{"error":"invalid_key"}`.

The UI shows "Detail content isn't available for this item yet", subtotal Pending, Book disabled.

## Root cause

`GET /v1/public/cruises/:id/content` (`getCruiseContent` in `@voyant-travel/cruises`) serves **sourced** cruises only — it reads a catalog sourced-entry row (`readSourcedEntry("cruises", id)`) and returns `null` (→404) when there is none. Unlike products, which have `buildOwnedProductContent` (`packages/inventory/src/service-content-owned.ts`) and serve owned + sourced from the same endpoint, cruises have **no owned-cruise content projector**.

The operator seeds **owned** cruises (`source.kind = "owned"`, `cru_*` ids) and indexes them into customer search (`cruiseRowToProjection`), so those cards linked to a detail page that can never render. Non-sourced demo ids (`cdmi_*`, multiple underscores) additionally fail `parseUnifiedKey` → 400.

## Chosen option + why

**Option 2 — stop search from surfacing unrenderable cruises**, by removing `cruises` from `storefrontCustomerBookableProductVerticals`.

Option 1 (make the resolver accept these ids) is **not a bounded fix**: the owned-cruise 404 is not a key-parsing problem (`cru_*` already parses with `allowOwnedKeys: true`) — it needs a full owned-cruise content projector (a ~500-line mirror of `buildOwnedProductContent`, projecting cruise + ship + sailings + cabin categories + itinerary + policies), which is a feature, not a bug fix, and higher risk. Normalizing the `cdmi_*` id would only turn a 400 into a 404.

This mirrors the exact pattern established for charters/flights in #2640/#2704: search derives its accepted verticals from `storefrontCustomerBookableProductVerticals`, so gating the list keeps search, the search-schema, and the customer detail route in sync automatically. Cruises stay fully searchable in the catalog and admin-manageable; only the customer storefront stops linking to them.

## Files changed

- `packages/storefront-react/src/routing.ts` — drop `"cruises"` from `storefrontCustomerBookableProductVerticals` (+ rationale comment). `isStorefrontCustomerBookableProductVertical("cruises")` now returns `false` and `getStorefrontCustomerProductDetailRoute("cruises", …)` returns `null`.
- `packages/storefront-react/src/routing.test.ts` — assert cruises are excluded.
- `starters/operator/src/routes/(storefront)/shop.tsx` — remove the cruises `<option>` from the vertical select; update the gating comment.
- `starters/operator/src/routes/(storefront)/shop.test.ts` — assert `?vertical=cruises` degrades to the default vertical.
- `starters/operator/src/routes/(storefront)/shop_.products.$entityModule.$entityId.tsx` — remove the now-unreachable cruise detail branch (the bookable-vertical guard above it made `entityModule === "cruises"` a TS no-overlap error).
- `.changeset/storefront-cruise-detail-content.md` — patch `@voyant-travel/storefront-react`.

The `CruiseDetailPage` component (`shop-product-detail-cruises.tsx`) and its test are intentionally retained so cruises can be re-enabled with a one-line change once owned cruises render public content end-to-end.

## Deferred

- Owned-cruise public content projector (`buildOwnedCruiseContent`) so cruises can be re-added to the bookable verticals. This is the "cruises stay shoppable" outcome, but out of scope for this fix.

Fixes #2639

### DIFF
--- a/.changeset/storefront-cruise-detail-content.md
+++ b/.changeset/storefront-cruise-detail-content.md
@@ -1,0 +1,12 @@
+---
+"@voyant-travel/storefront-react": patch
+---
+
+Stop the storefront customer surface from linking to cruise detail pages that
+cannot render. The public cruise content endpoint serves sourced cruises only
+(no owned-cruise content projector), so owned/demo cruises surfaced by catalog
+search linked to a detail page that 404s or 400s. Cruises are removed from
+`storefrontCustomerBookableProductVerticals` so search and the customer detail
+route no longer offer them, mirroring the charter/flight gating from
+voyant#2640. Cruises remain fully searchable and admin-manageable; re-add the
+vertical once owned cruises can render public content end-to-end.
--- a/packages/storefront-react/src/routing.test.ts
+++ b/packages/storefront-react/src/routing.test.ts
@@ -8,16 +8,20 @@ import {
 
 describe("storefront customer product routing", () => {
   it("keeps the customer detail route limited to implemented booking verticals", () => {
-    expect(storefrontCustomerBookableProductVerticals).toEqual([
-      "products",
-      "cruises",
-      "accommodations",
-    ])
+    expect(storefrontCustomerBookableProductVerticals).toEqual(["products", "accommodations"])
   })
 
   it("treats charters as searchable but not customer-bookable through product detail", () => {
     expect(isStorefrontCustomerBookableProductVertical("charters")).toBe(false)
     expect(getStorefrontCustomerProductDetailRoute("charters", "chrt_123")).toBeNull()
+  })
+
+  it("excludes cruises from customer-bookable product detail (voyant#2639)", () => {
+    // The public cruise content endpoint serves sourced cruises only, so owned
+    // seeded cruises produced dead detail pages. Cruises stay searchable in the
+    // catalog but must not link to the storefront customer detail route.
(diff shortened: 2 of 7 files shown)


mihaipxm:
Thanks — addressed both:

- **P2 (direct booking link):** fixed in 588ffcd64. Added a `beforeLoad` guard to `/shop/book/$entityModule/$entityId` that redirects any non-bookable vertical back to `/shop`, using the same `isStorefrontCustomerBookableProductVertical` source of truth as search + the detail route — so a stale/bookmarked `/shop/book/cruises/:id` URL can't reach the broken reserve flow.
- **P1 (commit attribution):** false positive. This commit and its parent are authored *and* committed as `Mihai Ungureanu <mihai@pixelmakers.com>` (`git log -1 --format='%an <%ae> | %cn <%ce>'`) with no model name in the subject/body/trailers. The reported `Codex <codex@openai.com>` identity isn't in the actual VCS metadata.

@codex review

PR #2712, opened 2026-07-01
fix(booking-journey): block checkout on unpriceable quotes + surface reserve failures

mihaipxm:
## Root cause
The owned accommodation booking handler (`packages/accommodations/src/booking-engine/handler.ts`) returns a **settled** quote with `available: true`, `invalidReason: "rates_missing"`, and no pricing when the selected stay has no applicable rate plan. The journey's commit gate only blocked on a thrown `quote.error` or `available === false`, so this settled-but-unpriceable quote passed `canAdvance`/`canCommit`/`onConfirm`: the buyer could accept the sales contract and click Confirm, `POST /v1/public/catalog/book` returned **502 `RESERVE_FAILED`**, and the storefront handler only `console.error`'d it — dropping the customer back on Review with no visible error.

## Core fixes
1. **Block checkout on unpriceable quotes** (`packages/bookings-react/.../booking-journey.tsx`): derive `quoteUnpriceable` (a settled quote that is `available === false` or carries any `invalidReason`, e.g. `rates_missing`) and gate `canAdvance`, `canCommit`, `onConfirm`, and the Review Confirm (`canConfirm`) on it. Show a distinct "adjust your selection" banner **without** a retry button (a re-quote yields the same reason).
2. **Surface reserve failures**: `handleAccepted` now catches checkout-handler throws and surfaces them via `confirmError`; the storefront `defaultCheckoutHandler` throws localized errors on `/book` failure (distinct reserve-failed copy for `rates_missing`), missing `bookingId`, and `/checkout/start` error, and rethrows in the outer catch — so `<BookingJourney/>` renders a visible error instead of silently returning to Review.

## i18n / tests
- Added `pricingUnavailable`/`checkoutFailed` (bookings-react journey) + `checkoutFailed`/`reserveFailed` (operator storefront), en + ro; `pnpm i18n:check` passes.
- New `booking-journey.rates-missing.test.tsx`; all 85 bookings-react journey tests pass.
- Changeset: `@voyant-travel/bookings-react` patch.

## Secondary / deferred (from the issue)
- Auto-selecting the first room that has an applicable rate plan, and showing room/rate **names** (not raw IDs) in Review — deferred as follow-up UX.
- The React `Cannot update a component (BookingJourney) while rendering (PaymentStep)` warning is already fixed (intent-snap runs in a `useEffect`).

Fixes #2638

### DIFF
--- a/.changeset/quick-rooms-attend.md
+++ b/.changeset/quick-rooms-attend.md
@@ -1,0 +1,14 @@
+---
+"@voyant-travel/bookings-react": patch
+---
+
+Block booking commit on un-priceable quotes and surface checkout failures.
+
+The booking journey now treats a settled quote that reports an `invalidReason`
+(e.g. the owned accommodation handler's `rates_missing`) or is explicitly
+unavailable as un-priceable: Next, contract acceptance, and Confirm are gated
+and a clear "adjust your selection" message is shown, instead of letting the
+buyer commit an unpriced booking that fails with a 502 `RESERVE_FAILED` at
+`/book`. A checkout handler that throws (e.g. the storefront `/book` +
+`/checkout/start` flow) now renders a visible error in the checkout UI rather
+than dropping the customer back on Review with only a console log.
--- a/packages/bookings-react/src/i18n/en-journey.ts
+++ b/packages/bookings-react/src/i18n/en-journey.ts
@@ -37,6 +37,9 @@ export const bookingsUiEnJourney = {
       retryQuote: "Retry pricing",
       quoteUnavailable:
         "Pricing isn't confirmed yet — retry pricing above before confirming your booking.",
+      pricingUnavailable:
+        "This selection can't be priced right now. Please adjust your choice before confirming your booking.",
+      checkoutFailed: "We couldn't complete your booking. Please try again.",
       addAtLeastTravelers: "Add at least {count} traveler{plural} to continue.",
       maxTravelersPerBooking: "Max {count} travelers per booking.",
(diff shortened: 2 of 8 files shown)


mihaipxm:
Fixed in 1a94e4188. `canAdvance` now hard-blocks only on a thrown `quote.error` (transient — a retry fixes it). An un-priceable quote (`rates_missing`) is corrected by *navigating* to the room/rate editor, so Next stays enabled and the shopper can reach the corrective step. Commit stays fully gated on `quoteBlocked` (`canCommit`, `onConfirm`, and the Review `canConfirm`), so an unpriced booking still can't be submitted, and the 'adjust your selection' banner still shows. 85 journey tests pass.

@codex review

mihaipxm:
Fixed in 605031b09. The RESERVE_FAILED body nests the reason under `context.upstreamPayload.reason` (the engine's `ReserveFailedError` serialized via the standard `{error, code, context}` envelope), which matches the exact payload in the issue. Changed the lookup to `errBody.context?.upstreamPayload?.reason` so `rates_missing` reserve failures now show the 'adjust your selection' copy instead of the generic message.

@codex review

mihaipxm:
Fixed in b9cbb4a05. Added a `CheckoutError` marker class for our intentional, already-localized checkout throws; the outer catch now only preserves `CheckoutError` and wraps everything else (native fetch/network error, `Response.json()` parse of an HTML 502) in the generic checkout message — so raw browser/parser text never reaches the shopper.

@codex review

PR #2719, opened 2026-07-01
feat(distribution): emit product.publication.changed for catalog reindex on channel mapping changes

mihaipxm:
Fixes #2636

## Root cause

Product/content/availability/pricing/promotion changes already drive catalog
reindexing via the operator catalog bridge's event subscribers, but the
distribution product↔channel **mapping** mutations emitted **no event**. So
adding an active product to an active channel made it pass the storefront
listability predicate (`hasActiveSalesChannelMapping` → active mapping to an
active channel), while the customer search collection stayed stale until some
other product mutation or a manual reindex ran. The inverse
(deactivate/remove) could not tombstone the slice promptly either.

## The event

`product.publication.changed` (dot-case `<resource>.<pastTenseAction>`, matching
the existing event vocabulary — `availability.slot.changed`,
`pricing.rule.changed`, `promotion.changed`). Payload:

- `productId`, `channelId`, `mappingId`
- `previousActive` / `nextActive` — the mapping active state before/after
  (`null` = didn't exist / removed)
- `operation` — `created | updated | deleted | activated | deactivated`
  (activate/deactivate are the update cases where `active` flipped)
- `channelKind` / `channelStatus` — channel kind/status at emit time (diagnostic)

Emitted from the **service layer** (`emitProductPublicationChanged` in
`packages/distribution/src/events.ts`), so every write path is covered — the
`batch-update` / `batch-delete` routes fan out over the same single-item
`updateProductMapping` / `deleteProductMapping` service methods, which now
receive the request-scoped bus and emit per id. Emission is fire-and-forget and
never throws, per the EventBus contract.

## The subscriber (packaged default)

The operator catalog bridge (`starters/operator/src/api/subscribers/catalog-bridge.ts`)
subscribes to `product.publication.changed` and reindexes the affected
product's customer-facing slices (`reindexEntity("products", productId)`) —
mirroring the existing `availability.slot.changed` / `pricing.rule.changed`
subscribers exactly. It re-derives listability from current DB state (the
single-mapping payload is only a trigger), so publish (mapping added/activated)
and unpublish (deactivated/deleted) both re-evaluate the storefront gate. The
bridge is already wired into `app.ts`, so no additional deployment wiring is
needed.

## Files changed

- `packages/distribution/src/events.ts` (new) — event const, payload type, emit
  helper, `classifyMappingUpdate`
- `packages/distribution/src/service/commercial.ts` — emit on create/update/delete
- `packages/distribution/src/routes/env.ts` + `routes.ts` — thread `eventBus`
  through single-item + batch product-mapping handlers
- `packages/distribution/src/index.ts` — export the event surface
- `starters/operator/src/api/subscribers/catalog-bridge.ts` — reindex subscriber
- Tests: `routes.product-mappings.events.test.ts` (service-layer emission incl.
  batch), `catalog-bridge.publication.test.ts` (subscriber reindex trigger)
- Changeset (minor) for `@voyant-travel/distribution`

## Verification

- `@voyant-travel/distribution` typecheck + biome clean; subscriber test passes
  (2/2). Service-layer DB integration tests are `describeIfDb`-gated and run in
  CI (no local Postgres/Docker in the sandbox). Pre-commit + pre-push full turbo
  build/typecheck (96/96) passed.

### DIFF
--- a/.changeset/distribution-publication-event.md
+++ b/.changeset/distribution-publication-event.md
@@ -1,0 +1,21 @@
+---
+"@voyant-travel/distribution": minor
+---
+
+Emit `product.publication.changed` on every product↔channel mapping mutation.
+
+The distribution service now emits a durable `product.publication.changed`
+domain event from the service layer whenever a channel product mapping is
+created, updated, deleted, activated, or deactivated — including the
+batch-update / batch-delete paths, which fan out over the same single-item
+service methods. The payload carries `productId`, `channelId`, `mappingId`,
+the previous and new mapping active state, the operation source
+(`created | updated | deleted | activated | deactivated`), and the channel
+`kind` / `status` at emit time.
+
+This lets catalog / storefront integrations reindex a product's
+customer-facing slices the moment its publication changes (adding an active
+mapping to an active channel makes it listable; deactivating or removing one
+should tombstone the slice), instead of waiting for an unrelated product
+mutation or a manual reindex. Emission is fire-and-forget and never throws,
+per the EventBus contract.
--- a/packages/distribution/src/events.ts
+++ b/packages/distribution/src/events.ts
@@ -1,0 +1,112 @@
+/**
+ * Distribution domain events.
+ *
+ * Emitted (after the write commits) by every service path that mutates a
+ * product↔channel mapping — create, update, delete, and the activate /
+ * deactivate cases of update. Batch/import paths route through the same
+ * single-item service methods, so they emit too.
+ *
+ * `product.publication.changed` is the durable signal that a product's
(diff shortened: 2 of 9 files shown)


mihaipxm:
Fixed in a07b23d69. `updateProductMapping` now captures the prior `productId`/`channelId` and, when a PATCH reassigns the mapping to a different product (or channel), emits a second `product.publication.changed` for the **old** (product, channel) with `operation: "deleted"`, `nextActive: null` — so the previous product is reindexed and can be tombstoned instead of staying stale. The subscriber re-derives listability from DB state, so this is a removal trigger. Added an integration test asserting a reassignment emits for both products.

@codex review

PR #2727, opened 2026-07-01
fix(bookings): expose child collections in admin booking detail read (refs #2620)

mihaipxm:
## Summary

Investigation of #2620 (anonymous storefront bank-transfer checkout of a connected product). The issue bundles several symptoms; this PR ships the one coherent, verifiable data-exposure fix and files a focused follow-up (#2723) for the deferred checkout-flow gaps, which turned out to be by-design rather than broken.

Using **Refs #2620** (not "Fixes") because this resolves the admin-API child-record exposure but intentionally defers the post-checkout flow gaps to #2723.

## What was fixed

**Admin booking detail returned no child collections.** `GET /v1/admin/bookings/{id}` returned only the flat booking row. `items`, `travelers`, and `documents` exist in the database but were reachable exclusively through the per-collection sibling endpoints (`/{id}/items`, `/{id}/travelers`, `/{id}/documents`), so a client consuming the detail response alone saw them as `null`/absent.

- The detail response now composes the bookings-owned child collections (`items`, `travelers`, `documents`) via a dedicated `bookingDetailSchema`, reusing the existing `listItems`/`listTravelers`/`listDocuments` service methods.
- Traveler PII follows the same reveal/redaction gate as the standalone travelers read.
- **Finance-owned records (payments, invoices) are intentionally not inlined** — bookings must not depend on finance (module dependency runs finance→bookings). They stay served by the finance booking-scoped admin routes and are composed at the deployment boundary.

## What was investigated and NOT changed (with rationale)

- **"Finance invoices list exposes `number`/`kind`/`totalAmountCents` as null."** These field names do not exist in the codebase. The finance invoices list correctly exposes `invoiceNumber`, `invoiceType`, and `totalCents` (mapped from `invoice_number`/`invoice_type`/`total_cents`), and `@voyant-travel/finance-react` consumes those exact names — renaming would break the operator UI. This is a client-field-name mismatch in the reporter's ad-hoc query, not a serializer bug. No change.
- **`payments`/`invoices` on the booking detail** — deferred to the finance booking-scoped routes for the boundary reason above.

## Deferred to follow-up #2723 (by-design, needs product decision)

All of the "missing rows" trace to a single root cause: the storefront **bank-transfer** checkout completes synchronously only up to `awaiting_payment` (booking + proforma + draft contract + pending payment session). Payment-schedule persistence, contract numbering/PDF generation, and activity-log emission are deferred to the `payment.completed` → `checkout-finalize` workflow and `booking.confirmed` subscribers, which fire only after an operator marks the transfer received.

- `booking_payment_schedules` — persistence exists but is a `booking.confirmed` subscriber; the preview deposit/balance is computed and not persisted.
- `contracts.contract_number` blank — numbering happens at issue time (`booking.confirmed`); the blank-numbered draft is the current pre-payment artifact.
- `booking_activity_log` — no rows emitted at checkout-start.
- `booking_documents` — only the manual admin upload creates these; not part of any checkout path (absence expected).

## Files changed

- `packages/bookings/src/routes-admin.ts` — new `bookingDetailSchema`; detail route + handler compose items/travelers/documents.
- `packages/bookings/tests/admin/unit/routes-contract.test.ts` — contract tests asserting the detail schema hydrates the child collections and requires them (they can never be null/absent).
- `.changeset/booking-detail-child-collections.md` — patch changeset for `@voyant-travel/bookings`.
- Regenerated OpenAPI artifacts: `packages/openapi/spec/*.json`, `starters/operator/openapi/*.json`.

## Verification

- `@voyant-travel/bookings` typecheck + tests pass (contract test file: 17 passed).
- Pre-push hook (full turbo typecheck/build/test across the monorepo) green.

Refs #2620. Follow-up: #2723.

### DIFF
--- a/.changeset/booking-detail-child-collections.md
+++ b/.changeset/booking-detail-child-collections.md
@@ -1,0 +1,17 @@
+---
+"@voyant-travel/bookings": patch
+---
+
+Hydrate the bookings-owned child collections in the admin booking detail read.
+
+`GET /v1/admin/bookings/{id}` now returns the booking together with its `items`,
+`travelers`, and `documents` arrays instead of the flat booking row alone. These
+records previously existed in the database but were only reachable through the
+per-collection sibling endpoints, so clients that consumed the detail response
+on its own saw the nested collections as `null`/absent. Traveler PII continues to
+follow the same reveal/redaction gate as the standalone travelers read.
+
+Finance-owned records (payments, invoices) are intentionally not inlined here to
+respect the module boundary (bookings must not depend on finance); they remain
+served by the finance booking-scoped admin routes and are composed at the
+deployment boundary.
--- a/packages/bookings/src/routes-admin.ts
+++ b/packages/bookings/src/routes-admin.ts
@@ -1674,6 +1674,22 @@ const bookingDocumentSchema = z.object({
   createdAt: isoTimestamp,
 })
 
+// The booking detail read (`GET /{id}`) hydrates the bookings-owned child
+// collections inline so a single fetch exposes the records that exist for the
+// booking. Previously the detail response carried only the flat booking row and
+// these collections were reachable exclusively through the per-collection
+// sibling endpoints (`/{id}/items`, `/{id}/travelers`, `/{id}/documents`),
+// which made them read as `null`/absent to clients that consumed the detail
+// response alone. Finance-owned records (payments, invoices) are intentionally
+// NOT inlined here: bookings must not depend on finance (the module dependency
+// runs finance→bookings), so those stay behind the finance booking-scoped admin
+// routes and are composed at the deployment boundary.
+const bookingDetailSchema = bookingSchema.extend({
+  items: z.array(bookingItemSchema),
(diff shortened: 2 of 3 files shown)


mihaipxm:
Valid — the client parser does strip the newly-exposed collections. Deferring it to a focused follow-up (#2728) rather than fixing it here, for two reasons:

1. **No current consumer** reads `booking.travelers`/`booking.documents` from `useBooking` (the operator UI uses the per-collection hooks; the only `.travelers`/`.documents` reads are of the *draft* in the contract-variable builder). So the server exposure is non-breaking and there's no user-facing gap from the stripping today.
2. A correct client `bookingDetailSchema` is a fiddly, separable change with regression risk: travelers have a reveal/redact duality (base vs reveal schema with required `travelDetails`), and there is no client booking-level document schema yet — a wrong shape would make `fetchWithValidation` *throw* and break the detail page, which is worse than the current strip.

This PR is `Refs #2620` (the server-side null-collections bug is the reported issue and is fixed + contract-tested); #2728 tracks the client-parser preservation.

PR #2740, opened 2026-07-01
fix(inventory): link a customer from billing contact for anonymous owned-product booking

mihaipxm:
## Root cause
Anonymous storefront checkout for an **owned** public product failed at `POST /v1/public/catalog/book` with `400 {"error":"Select a billing person or organization","fieldErrors":{"personId":[...]}}`. The owned booking commit required a CRM `personId`/`organizationId`, which an anonymous B2C customer doesn't have — whereas the sourced/session arm creates a customer person from the billing contact. So owned products were unbookable anonymously.

## Fix
`createProductsBookingHandler` (`@voyant-travel/inventory`) gains an optional **`resolveBillingPerson`** bridge: when a commit carries no `personId`/`organizationId` but has a billing contact, it resolves/creates a CRM person from that contact and uses it as the booking's billing person. The operator wires the bridge to `relationshipsService.upsertPersonFromContact`, mirroring how the sourced/session arm links a customer. It's an **injected bridge**, so `@voyant-travel/inventory` takes no direct `@voyant-travel/relationships` dependency (module boundary preserved). No-op when a person/org is already supplied (operator/trips paths) or no resolver is wired.

- `packages/inventory/src/booking-engine/handler.ts` — `ResolveOwnedBillingPerson` type + resolve-on-missing-party logic
- `packages/inventory/src/booking-engine/index.ts` — export the new type
- `starters/operator/src/api/lib/product-booking-handler.ts` — wire `resolveBillingPerson` → `upsertPersonFromContact`
- `packages/inventory/tests/unit/booking-engine-quote.test.ts` — tests (anonymous contact → person resolved/linked; person/org already supplied → no-op)
- Changeset: `@voyant-travel/inventory` patch

## Secondary items (from the issue)
- **Error surfacing** — already handled by #2712 (merged): the storefront `defaultCheckoutHandler` throws a visible `CheckoutError` on `/book` failure, surfaced via the journey's `confirmError`. With this fix the 400 no longer occurs; if it did, it's now visible.
- **Missing bank-transfer/inquiry payment methods for owned products** — a separate payment-capability-resolution difference; not addressed here (candidate follow-up).

Fixes #2618

### DIFF
--- a/.changeset/anon-owned-product-booking.md
+++ b/.changeset/anon-owned-product-booking.md
@@ -1,0 +1,12 @@
+---
+"@voyant-travel/inventory": patch
+---
+
+Owned product booking commit now resolves (or creates) a CRM person from the
+billing contact when the commit carries no `personId`/`organizationId` — the
+anonymous storefront checkout case. `createProductsBookingHandler` accepts a new
+optional `resolveBillingPerson` bridge (wired by the template to
+`relationshipsService.upsertPersonFromContact`), mirroring the sourced/session
+arm's `resolveBillingPerson` hook. This fixes anonymous storefront checkout for
+owned public products, which previously failed with a 400 "Select a billing
+person or organization".
--- a/packages/inventory/src/booking-engine/handler-support.ts
+++ b/packages/inventory/src/booking-engine/handler-support.ts
@@ -364,6 +364,23 @@ export function extractBillingParty(party: Record<string, unknown> | undefined):
   }
 }
 
+// Mirrors `isRealEmail` in @voyant-travel/finance's `requireCompleteBookingParty`
+// (and the trips copy). The owned booking handler resolves a CRM person from the
+// billing contact before calling `createBooking`, which rejects a blank or
+// placeholder email — so the resolver must apply the same rule up front, or it
+// orphans a CRM person on every failed checkout. Keep this set in sync with
+// finance's `placeholderEmails`.
+const placeholderBillingEmails = new Set([
+  "noreply@example.com",
+  "tbd@example.com",
+  "traveler@example.com",
+])
+
+export function isRealBillingEmail(value: string | null | undefined): value is string {
+  const normalized = value?.trim().toLowerCase() ?? ""
+  return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(normalized) && !placeholderBillingEmails.has(normalized)
+}
+
 export function extractPartyTravelers(
   party: Record<string, unknown> | undefined,
 ): Array<{ personId?: string | null }> {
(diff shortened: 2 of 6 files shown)


mihaipxm:
Fixed in ba1407bf5. The resolver now only fires when the billing contact has a real contact point (email or phone). A name-only contact skips it — resolving on name-only would create a CRM person that still fails `createBooking`'s email/phone requirement, orphaning a person per attempt. Added a name-only-skip test (asserts the resolver isn't called and `personId` stays null so the commit rejects as before).

@codex review

mihaipxm:
Addressed in ed7e78e91: the owned booking handler now reads the billing contact from `request.party` **or** the saved `draft.billing.contact`, so the anonymous storefront path (which POSTs only a draftId) stamps the booking contact and resolves a CRM customer instead of 400ing. Added a draft-billing unit test. @codex review

mihaipxm:
Addressed in b4d1dec5b: billing contact points are now trimmed to null before the resolver gate, so a whitespace-only email/phone (including the draft schema's default `""`) is treated as absent and never triggers a name-only CRM person. Added a whitespace-only skip test. @codex review

mihaipxm:
Addressed in 13beb107e: the handler now applies the same `isRealEmail` rule `createBooking`'s `requireCompleteBookingParty` uses (new `isRealBillingEmail` helper mirroring the finance/trips copies, incl. the `placeholderEmails` set). A placeholder/malformed billing email is cleared to null before the resolver gate, so it never creates a CRM person that `createBooking` would reject — the resolver falls back to a real phone when present. Added a placeholder-email-with-phone test. @codex review

mihaipxm:
Also hardened in f1ce226ff: the resolver now only fires when the billing contact fully satisfies createBooking's person requirements — BOTH a first and last name AND a real email or phone — closing the remaining orphan-row path (contact point present but no name). @codex review

mihaipxm:
Addressed in 6896defe7: the resolver no longer stamps the provisional `request.bookingId`. The booking number is now generated up front and used as the CRM person's provenance ref (`source_ref`) — it's caller-supplied, written straight to the booking row (`booking.bookingNumber`), and known before create, so it resolves to the real persisted booking instead of the id the finance bridge discards. Test asserts the resolver receives the booking number, not `request.bookingId`. @codex review

mihaipxm:
Addressed in f72b97f87: the resolver gate now mirrors the full anonymous-path accept criteria of createBooking's requireCompleteBookingParty — it only fires when there's a complete billing person (name + real contact point) AND at least one traveler with a name (or linked person) and no placeholder email. So a draft that would fail on travelers no longer creates a CRM person first. The operator/trips paths supply a person/org and skip the block entirely; voucher/price-override rejections come from operator-only draft fields absent on the anonymous storefront. Added a no-travelers skip test. @codex review

PR #2766, opened 2026-07-01
feat(openapi): generate per-module specs, stop committing giant aggregates

mihaipxm:
Closes #2733.

## The problem

The admin/storefront OpenAPI surface was published as one multi-megabyte document — `framework-admin.json` was **7.3 MB**, which GitHub flat-out refuses to render ("we can't show files that are this big") and which was rewritten wholesale on any route change (unviewable PR diffs, git bloat). Nothing serves or imports it at runtime; it existed only as a drift golden.

## The approach

Rather than *build one giant doc → split it by path prefix*, **invert the flow**: generate one self-contained spec **per module, from the routes that module registered**, and let the deployment compose the aggregate on demand instead of carrying it in git.

The module boundary is taken from the composition's own mount manifest — authoritative, not a `/v1/admin/<seg>` guess. This matters for the storefront, where `publicPath` overrides mount routes under prefixes that aren't the module name (`/v1/public/booking-engine`, `/v1/public/customer-portal`); a path-prefix split would misattribute them.

## Changes

- **`packages/hono`** — new `generateModuleOpenApiDocuments(mounts, options)` in `@voyant-travel/hono/openapi`, plus an `app.moduleMounts` manifest (mirrors the existing `app.lazyMounts`) recorded in the mount loop. Each mount carries `{ moduleName, prefix, load }` (eager routes wrapped as `() => routes`; lazy pass their loader). Build-time only — stays out of the Worker bundle. Unit-tested (`openapi-modules.test.ts`).
- **`packages/openapi`** (published contract) — commits compact, browsable per-module specs under `spec/{admin,storefront}/<module>.json`, exposed via new `./admin/*` / `./storefront/*` subpath exports (e.g. `import bookings from "@voyant-travel/openapi/admin/bookings"`). The 7–9 MB aggregates are **gitignored** and generated at **`prepack`**, so the npm tarball still ships them (`.`/`./admin`/`./storefront` exports unchanged). The framework generator now also merges lazy families into `full` (a latent gap — the operator generator already did).
- **`starters/operator`** — same per-module split; giant aggregates gitignored and generated on demand via `generate:openapi`.

Both drift gates now assert the per-module files (content **and** the file set, so a removed module drops its stale file) plus that every per-module path exists in the aggregate (no invented routes).

## Net effect

- `framework-admin.json` 7.3 MB → per-module `admin/*.json` (largest `finance.json` ≈ 1.1 MB, most far smaller) — each independently viewable and diffable on GitHub.
- A bookings route change now rewrites `bookings.json` (596 KB), not the whole 7.3 MB admin file — localized diffs.
- ~14 MB of unviewable JSON leaves git; the aggregate is composed on demand (prepack for the published package; `generate:openapi` for the operator).

## Verification

- `pnpm --filter @voyant-travel/hono {typecheck,lint,test}` — 256 tests pass (incl. new helper unit tests).
- `pnpm --filter @voyant-travel/openapi {typecheck,lint,test}` + `pnpm --filter operator {typecheck,test}` — drift gates green.
- `pnpm pack` from a clean state (giants deleted) — `prepack` regenerates them; tarball contains `framework-*.json` **and** `spec/{admin,storefront}/*.json`.
- Regeneration is deterministic (re-running produces no per-module diff); giants confirmed gitignored.
- `pnpm release:plan` — clean.

## Out of scope (follow-ups)

- The Swagger/Scalar **docs UI** with a module selector (the issue frames it as "could then…"). The per-module artifacts are the enabling step.
- `x-voyant-module` / `x-voyant-surface` metadata (#2729) — now trivial to stamp since the generator knows the owning module per operation.

Note: the sibling per-module split for the operator drops the previous single committed aggregate golden; drift protection is preserved at finer (per-module) granularity.

Pre-existing failures on `main` (commerce `service-price-schedules.test.ts` typecheck, bookings-react `booking-contract-dialog` typecheck) are unrelated to this change.


mihaipxm:
Thanks @chatgpt-codex-connector — both P2s addressed in `847daafeb`:

**Preserve non-module routes (workflow-runs).** Correct — the per-module specs were built only from `app.moduleMounts`, dropping `additionalRoutes` mounts like the operator's `/v1/admin/workflow-runs/*`. Added `splitDocumentByModule`, which partitions the *full* document so every `/v1/admin/*` and `/v1/public/*` path lands in exactly one module file — the manifest is the authoritative owner (keeps `publicPath` overrides correct), and the path's own segment is the fallback for routes no module claims. `starters/operator/openapi/admin/workflow-runs.json` is now committed, and a new completeness test asserts `⋃ per-module surface paths == aggregate surface paths` so nothing can be silently dropped again. (`_meta/capabilities` is a plain `.get()` route with no `.openapi()` registration, so it was never in any generated spec — not a regression.)

**Keep aggregate exports resolvable.** Fair — the `.`/`./admin`/`./storefront` exports pointed at gitignored aggregates absent in a fresh checkout. Added a `build` script (`pnpm generate`, wired uncached after `^build` in `turbo.json`) so `pnpm build` regenerates them locally; `prepack` still covers the npm tarball. The aggregate exports now resolve after the standard build, without carrying the 7 MB files in git.

PR #2851, opened 2026-07-03
Fix sourced package booking path

mihaipxm:
Fixes #2847.

## Summary
- Preserve sourced Connect provenance through quote/book adapter context.
- Map sourced package checkout data into Connect package confirm parameters, including travelers, lead traveler, contact, and `connectRoute: "packages"`.
- Add an operator prepare hook that creates a Connect package lock from the quoted offer before booking.
- Send storefront checkout party details to public catalog booking.

## Coordination
- Depends on connect#101 shipping with this change; the published Connect adapter does not yet route `connectRoute: "packages"` reserves to the package confirmation path.
- ProTravel does not consume this starter runtime. After package publish, ProTravel needs a follow-up to add the equivalent `prepareBookParameters` hook in its catalog booking composition and bump the `@voyant-travel/catalog` / `@voyant-travel/connect-adapter` overrides.

## Verification
- `pnpm exec biome check ...` on touched files
- `pnpm --filter @voyant-travel/catalog typecheck`
- `pnpm --filter @voyant-travel/catalog build`
- `pnpm --filter @voyant-travel/catalog test -- routes.test.ts`
- `pnpm --filter operator test -- src/api/runtime/catalog-booking-runtime.package-hold.test.ts`
- `pnpm --filter operator test -- src/api/runtime/catalog-booking-runtime.package-hold.test.ts src/api/runtime/catalog-booking-runtime.test.ts`
- `pnpm --filter operator test -- src/components/voyant/booking-journey/storefront-booking-journey.test.ts src/api/runtime/catalog-booking-runtime.test.ts`
- `pnpm exec tsc -p tsconfig.server.json && pnpm exec tsc -p tsconfig.client.json` in `starters/operator`
- `pnpm verify:agent-quality:changed`
- Commit hook: 186/186 tasks successful
- Push hook: 98/98 tasks successful

### DIFF
--- a/.changeset/sourced-package-booking.md
+++ b/.changeset/sourced-package-booking.md
@@ -1,0 +1,5 @@
+---
+"@voyant-travel/catalog": patch
+---
+
+Thread sourced package booking drafts into Connect package quote/book parameters, including route, contact, and traveler details.
--- a/packages/catalog/src/booking-engine/index.ts
+++ b/packages/catalog/src/booking-engine/index.ts
@@ -201,6 +201,7 @@ export {
   type CatalogBookingHoldPlaceBody,
   type CatalogBookingHoldReleaseBody,
   type CatalogBookingHoldTtlInput,
+  type CatalogBookingPrepareBookParametersInput,
   type CatalogBookingProvenance,
   type CatalogBookingProvenanceInput,
   type CatalogBookingQuoteBody,
--- a/packages/catalog/src/booking-engine/routes-contracts.ts
+++ b/packages/catalog/src/booking-engine/routes-contracts.ts
@@ -8,6 +8,7 @@ import type { BookEntityResult } from "./book.js"
 import type { OwnedBookingHandlerRegistry } from "./owned-handler.js"
 import type { QuoteEntityDeps, QuoteEntityResult } from "./quote.js"
 import type { SourceAdapterRegistry } from "./registry.js"
+import type { SelectCatalogQuote } from "./schema.js"
 import type { SnapshotContentCapturer } from "./snapshot-content.js"
 
 const recordSchema = z.record(z.string(), z.unknown())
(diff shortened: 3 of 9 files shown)


mihaipxm (commented):
This is the right shape — it closes all three gaps from #2847 (prepare-hook hold creation, draft → `PackageConfirmInput` mapping incl. band→category and documents.sex/gender normalization, and `connectRoute` threading with provenance now riding the quote parameters too). Three findings from checking it against the live consumer path:

1. **Bug — `isPackageOffer` requires `offer.status`, which `PackageOffer` doesn't have.** Neither the connect-sdk type (`PackageOffer`: id/connectionId/supplierId/productRef/title?/stay/flights/pricing/cancellationPolicy/expiresAt) nor the platform contract (`packageOfferSchema`) carries a `status` field — that's `PackageHold`'s field. As written, `readPackageOffer` returns `undefined` for every valid stored offer, `prepareConnectPackageBookParameters` falls through without locking, and confirm then 400s on the missing `holdId` — i.e. the exact flow this PR adds never fires. Dropping the `offer.status` check (or moving it to an optional) fixes it.

2. **Landing order — the reserve branch this hands off to ships in voyant-travel/connect#101.** `connectRoute: "packages"` reserve → `packages.confirm` (plus cancel/status for `package:` refs) is in the adapter PR; on the current published `@voyant-travel/connect-adapter`, the reserve falls through to the generic `bookings.create` path. Worth cross-linking so deployments bump both together.

3. **Nit — `isConnectPackageCandidate` special-cases `sourceProvider === "tui"`.** The generic room/board-pin check right after it already covers that case, so the provider literal is redundant — and keeps supplier names out of the framework.

Also a non-blocking note: the stored quoted offer is short-TTL (per-search offer codes), so a slow checkout will lock an already-expired offer with no re-resolve fallback — fine as v1, might deserve a TODO to re-resolve with the roomTypeId/ratePlanId/board pin when the lock rejects.

PR #2974, opened 2026-07-06
Make Node the first-class production target for the operator

mihaipxm:
Closes #2966.

Makes **Node the first-class (and only) runtime for the operator**. A composed operator API cannot stay resident on Cloudflare Workers (evaluated-heap eviction → multi-second per-request graph evaluation); Node (a resident process, e.g. Cloud Run) serves it warm in 1–2 ms. Per the maintainer comment on the issue this is *node-only — no workers/node split* — while Workers stays supported for storefronts and federated per-domain apps, so `packages/hono`, `packages/framework`, and `packages/worker-runtime` are left runtime-neutral and untouched (they already degrade to KV/no-op on Node).

## What changed

**`@voyant-travel/dedicated-runtime`** — dropped the Cloudflare-emulation shims (KV-over-REST, the `caches.default` shim, `buildDedicatedEnv`); with no Workers shape to emulate they had no purpose. Added real Node providers: `createMemoryKvNamespace` (in-process KV for `CACHE`/`RATE_LIMIT`), `createMemoryR2Bucket` (dev object store), and `composeNodeEnv`. The S3-backed `createR2BucketShim` and the `node-server`/`trust`/`wait-until` survivors stay.

**Operator** — single Node entry `src/server.ts` boots `createNodeServer` (origin-trust gate, `/__voyant/scheduled` Cloud Scheduler hook, real `waitUntil` + graceful drain) and serves the client build; `src/entry.ts` now exports plain `fetch`/`scheduled` handlers. The Vite build drops `@cloudflare/vite-plugin` and targets Node (`ssr.target: "node"` + `noExternal` bundling of workspace packages, which use `.js`-extension specifiers Node can't resolve from source). `env.d.ts` retypes the bindings to real Node providers and drops `@cloudflare/workers-types`; `wrangler.jsonc` and `pg.d.ts` are removed.

**Config / crons / DB** — `deployment: "node"` (and `"dedicated"` added to `DeploymentTarget`, documented as the operator recommendation); a runtime-neutral cron table (`OPERATOR_CRON_JOBS`) plus `scripts/emit-cloud-scheduler.ts` that fans it out to Cloud Scheduler jobs; a resident pooled node-postgres DB lane via `DATABASE_URL_DIRECT` (`adapter: "node"`), with neon-http/WS kept as fallbacks.

[...]

PR #2979, opened 2026-07-06
Operator dev env: use .env (Node-native), drop wrangler .dev.vars

mihaipxm:
Follow-up on the Node-only operator (#2974 / #2976): fix the dev/local **env** story.

## Problem
The operator is Node-only, but its dev env still used the wrangler **`.dev.vars`** convention — and `src/server.ts` never loaded it (wrangler did, on Workers). So `pnpm dev` / `pnpm start` booted with **no secrets** and 500'd on "Missing BETTER_AUTH_SECRET". (This — not any routing bug — is what made `/api` look broken; the earlier README caveat was a mis-diagnosis.)

## Fix
- **`.env`, loaded the Node way.** `dev` and `start` run under `node --env-file-if-exists=.env` — no code loader, no dependency; already-set/platform env wins, and the file is absent in real deployments (no-op there).
- **`.dev.vars.example` → `.env.example`**, folding in the public vars the deleted `wrangler.jsonc` used to provide (`APP_URL`, `DASH_BASE_URL`, `EMAIL_FROM`, `VOYANT_ADMIN_AUTH_MODE`, `KMS_PROVIDER`).
- Operator scripts + `drizzle.config.ts` read `.env`; operator/root/demo-api READMEs updated. **`federated-operator` stays on `.dev.vars`** — it's still a Workers/wrangler deployment.
- Corrects the operator README dev note (dev serves `/api/*` fine).

## Verified
With `.env` present, `pnpm dev` → `/api/auth/get-session` **200**, `/api/openapi.json` **401** (auth), SSR `/` **200** — no missing-secret errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,8 @@ cd my-travel-app
 pnpm install
 
 # 3. Configure the app
-cp .dev.vars.example .dev.vars
-# set your secrets in .dev.vars, and DATABASE_URL in .env
+cp .env.example .env
+# set your secrets (DATABASE_URL, BETTER_AUTH_SECRET, …) in .env
 
 # 4. Run the app
 pnpm db:migrate
--- a/apps/catalog-demo-api/README.md
+++ b/apps/catalog-demo-api/README.md
@@ -20,7 +20,7 @@ pnpm db:migrate             # creates catalog_demo_inventory + catalog_demo_orde
 pnpm dev                    # listens on :3330 by default
 ```
 
-Then in your operator starter (e.g. `starters/operator/.dev.vars`):
+Then in your operator starter (e.g. `starters/operator/.env`):
 
 ```
 CATALOG_DEMO_API_URL=http://localhost:3330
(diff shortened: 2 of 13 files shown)


mihaipxm:
Addressed both:

- **P2 (env-file flag)** — replaced `--env-file-if-exists` (Node 22.9+/20.18+, and even 22.0–22.8 lack it) with a Node-native `-r ./scripts/env-preload.cjs` preload. `--require` works across the whole `>=20` range; the preload calls `process.loadEnvFile` when present (20.12+) and no-ops otherwise, so it never crashes and never overwrites already-set/platform env. Verified `pnpm dev` still loads `.env` (auth 200, no missing-secret).
- **P1 (explicit DATABASE_URL)** — `backfill-custom-fields.ts` and `reindex.ts` now capture an explicit `DATABASE_URL` before the `.env` override-load and restore it after (same guard `migrate.ts`/`drizzle.config.ts` use), so `DATABASE_URL=<target> tsx scripts/…` can't be redirected at a local `.env` DB.

PR #2981, opened 2026-07-06
Add a reference Dockerfile + Cloud Run recipe for the operator

mihaipxm:
The two nice-to-haves from the Node-first arc, as their own PR.

## Reference Dockerfile + Cloud Run recipe
- **`starters/operator/Dockerfile`** — multi-stage pnpm-monorepo build (build from the **repo root**: `docker build -f starters/operator/Dockerfile .`). The `@voyant-travel/*` packages are bundled into the server bundle; the runtime image ships the operator's production `node_modules` via `pnpm deploy --prod`.
- **`starters/operator/Dockerfile.dockerignore`** — BuildKit per-Dockerfile ignore (the repo-root `.dockerignore` is an orphaned artifact that excludes `packages/*`, which the workspace build needs).
- **README Deploy section** rewritten with the container-image steps + a **Cloud Run recipe** (Cloud Build → `gcloud run deploy --min-instances=1 --cpu-boost`, secrets via Secret Manager, `/healthz` probe, `emit:cloud-scheduler` wiring).

## Build simplification (enabled here)
The image build originally needed the whole dependency graph prebuilt (`turbo ^build`) because a few workspace subpaths (e.g. `@voyant-travel/workflows-orchestrator/in-memory`) export `node` → prebuilt `dist`. The operator Vite SSR config now puts **`development` ahead of `node`** in `ssr.resolve.conditions`, so those resolve to `./src` and get bundled — `pnpm --filter operator build` stands alone. The Dockerfile and the CI **`node-smoke`** lane drop `turbo ^build` accordingly (faster, lighter — no 7.7 GB dep-graph build).

## Also
- Move **`pg` → `dependencies`** (imported at runtime by `db.ts`, so `--prod` install must keep it; lockfile synced).
- Tidy a stale **`[worker-runtime]` log label → `[runtime]`** in `packages/runtime` (patch changeset).

## Verified
`docker build` → 850 MB image; `docker run` → `/healthz` **200**, SSR `/` **200**, `/api/openapi.json` **401** (auth dispatch), "[operator] Node runtime listening on :8080".


mihaipxm:
Thanks — one real, one false positive:

- **P2 (Cloud Build `--tag`/`--config`)** — correct, they're mutually exclusive. Replaced the recipe with a plain `docker build -f starters/operator/Dockerfile -t "$IMAGE" . && docker push "$IMAGE"` (the custom Dockerfile path rules out the bare `--tag` shortcut anyway). Fixed.

- **P1 (fresh-checkout build)** — this doesn't reproduce. The **`node-smoke`** CI job runs the exact `pnpm --filter operator build` on a fresh checkout (only `pnpm install` before it — no prebuilt `dist`), then boots the server and asserts `/healthz` 200: it's **green (1m15s)** on this PR head. `@voyant-travel/runtime`'s `.` export is a plain `./src/index.ts` string (always source, no conditions), and the other workspace subpaths resolve via `development`→`./src`, so `src/server.ts`'s imports bundle from source with no dep build. I verified the same locally by hiding `packages/workflows-orchestrator/dist` and building + booting successfully, and the Docker image (identical build step) builds and serves. No change needed there.

PR #3065, opened 2026-07-09
Gate source-free managed admin composition by the profile's module subset

mihaipxm:
## Summary

Closes #3063.

A managed profile can declare a **module subset** (`modules: [...]` in the snapshot), and the runtime honors it for the **API**. But the **source-free admin UI ignored it** — it composed *every* `create<Module>AdminExtension()` factory unconditionally, so every managed operator saw the full nav (~18 modules) even when its profile activated a subset. Nav entries for inactive modules linked to pages whose API isn't mounted → dead links / 404s.

This is inherently a **shared-image** problem: the managed admin is one framework-version-tagged image; the active-module set is per-operator, injected at deploy via the snapshot. The client bundle can't know the subset at build/import time, so the composition needs a **runtime signal**.

## What changed

**1. Runtime exposes the active module set** (`@voyant-travel/framework`)
- `resolveActiveModuleIds(project)` (`/profile`) — maps the resolved `include` set (the same one that drives `createVoyantApp({ exclude })`, so nav can never drift from what the API mounts) to `moduleId`s.
- `GET /auth/bootstrap-status` now returns `modules` on `ManagedBootstrapStatus` — the probe the workspace already issues at bootstrap.

**2. Admin gates composition by it** (`@voyant-travel/admin` + `starters/managed-operator`)
- `AdminBootstrapStatus` carries an optional `modules` list.
- The route tree + destinations stay built from the FULL registry (kept **hydration-stable** across the shared image); the **nav/widgets** are filtered at render (`WorkspaceContent`) via `filterManagedAdminExtensionsByModules`, keyed by `MANAGED_ADMIN_EXTENSION_MODULE_IDS` (extension id → required module id; `core` always active). Filtering is **fail-open**: if the runtime reports no set, everything is kept.
- The managed-operator dev `/auth` stub computes the same set from the snapshot, so subsets are exercisable locally.

`mice` maps to a module absent from the standard manifest, so gating drops it whenever the runtime reports its set — removing an extension whose API the managed runtime never mounts.

## Verification

- `pnpm verify:fast` green; framework + admin + managed-operator typecheck/lint/build.
- Unit tests: `resolveActiveModuleIds` (full set + subset + no-mice/no-storefront); `bootstrap-status` echoes the active modules; `filterManagedAdminExtensionsByModules` (gate / drop-mice / fail-open / empty-set → core-only).
- End-to-end data contract confirmed against the built Node server: default snapshot → full resolved set (no `mice`, no `storefront`); a `[bookings, catalog]` subset snapshot → `[action-ledger, relationships, identity, commerce, catalog, bookings]` (declared subset + required foundational), which through the filter yields nav **core, bookings, catalog, relationships, commerce, action-ledger** — everything else dropped.

## Scoped follow-up

Direct-URL navigation to a disabled module still resolves its route (page renders, then its API 404s). Per-route gating would need routes to carry an owning-module tag; nav gating fully resolves the reported symptom (dead links in the sidebar). Under a per-operator *build* the subset bakes in and no runtime signal is needed (see platform#1014).

### DIFF
--- a/.changeset/managed-admin-module-subset-gating.md
+++ b/.changeset/managed-admin-module-subset-gating.md
@@ -25,3 +25,9 @@ mounted (dead links / 404s).
 - `AdminBootstrapStatus` carries an optional `modules` module-id list the
   source-free admin filters its nav/widget composition by (fail-open when
   absent).
+- `AdminWorkspaceShell` accepts `activeModuleIds`; when provided, the packaged
+  **base** nav is gated to those modules (`filterAdminNavigationByModules` /
+  `OPERATOR_ADMIN_NAV_MODULE_IDS`), so a shared image's static nav (Flights,
+  Finance, Legal, …) is hidden for a profile subset — not just
+  extension-contributed items. Fail-open when omitted, so self-hosted starters
+  are unaffected.
--- a/.changeset/managed-admin-module-subset-gating.md
+++ b/.changeset/managed-admin-module-subset-gating.md
@@ -1,0 +1,27 @@
+---
+"@voyant-travel/framework": minor
+"@voyant-travel/admin": minor
+---
+
+Expose the managed profile's active module set at runtime so the source-free
+admin can gate its composition by the deployment's module subset (voyant#3063).
+
+The managed runtime already honors a profile's `modules: [...]` subset for the
+API (`createVoyantApp({ exclude })`), but the shared, framework-version-tagged
+admin image composed *every* `create<Module>AdminExtension()` factory
+unconditionally — so every managed operator saw the full nav even when its
+profile activated a subset, with nav entries linking to pages whose API isn't
+mounted (dead links / 404s).
+
+`@voyant-travel/framework`:
+
+- Add `resolveActiveModuleIds(project)` (`/profile`) — the resolved module `include`
+  set (the same one that drives `createVoyantApp({ exclude })`) as `moduleId`s.
+- `GET /auth/bootstrap-status` now returns `modules` on `ManagedBootstrapStatus`,
+  so the workspace bootstrap probe learns what's active for this deployment.
+
+`@voyant-travel/admin`:
+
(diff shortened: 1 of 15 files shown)


mihaipxm:
Good catch on the P1 — fixed in 040f3b2.

The gap was real: filtering `extensions` only dropped extension-contributed nav; the sidebar falls back to the static `createOperatorAdminNavigation` base list (catalog, flights, products, suppliers, finance, legal, …), which is independent of extension filtering. So under `modules: ["bookings", "catalog"]` the inactive base entries stayed visible and still linked to unmounted APIs.

**Fix:** gate the base nav in the packaged admin.
- `@voyant-travel/admin`: `AdminWorkspaceShell` now accepts `activeModuleIds` and threads it through `AdminWorkspaceShellInner` → `OperatorAdminWorkspaceLayout`, which filters the resolved nav via `filterAdminNavigationByModules` keyed by `OPERATOR_ADMIN_NAV_MODULE_IDS` (nav-item id → module id, incl. non-identity maps: `products`→`inventory`, `availability`/`resources`→`operations`, `suppliers`/`channel-sync`→`distribution`, `people`/`organizations`→`relationships`). Fail-open when omitted, so self-hosted starters are unaffected.
- `starters/managed-operator`: the host passes the fetched `activeModuleIds` to the shell.

**Verified in-browser** against the built Node server. `modules: ["bookings", "catalog"]` (resolved active set `[action-ledger, relationships, identity, commerce, catalog, bookings]`) collapses the sidebar from 18 sections to: **Dashboard, Catalog, Bookings, People, Organizations, Promotions (commerce), Logs (action-ledger), Settings** — Flights, Products, Availability, Notifications, Suppliers, Resources, Finance, Legal, Distribution, Trips, Quotes are all gone. Added unit coverage for the base-nav filter (real nav + non-identity mappings + fail-open + empty-set).

PR #3070, opened 2026-07-09
Export managed-profile scheduled-jobs + workflow manifests (no-build provisioning)

mihaipxm:
## Summary

Closes #3032.

Managed-profile deployments (platform#953/#954) run a fixed `voyant-operator-runtime:<framework-version>` image with **no build step** — the platform never sees a build artifact. Two Cloud provisioning steps needed data the framework didn't export, derivable purely from a profile snapshot:

- **Gap 1** — the Cloud Scheduler job set (`OPERATOR_CRON_JOBS` + `emit:cloud-scheduler` lived only in `starters/operator`).
- **Gap 2** — the workflow definitions manifest at `{ id, config }` grain (only reachable from a build artifact).

## What changed

New subpath **`@voyant-travel/framework/managed-jobs`**, both derived from the **resolved module subset** (like `resolveActiveModuleIds`), so a profile that activates fewer modules provisions fewer jobs/workflows:

- `getManagedProfileScheduledJobs(project)` → `{ id, cron, description, route, module }[]`. Always-on framework infra (`outbox-drain`) plus one job per active owning module: `draft-reaper`→catalog, `promotion-boundary-scheduler`→commerce, `channel-push-*`→distribution. Dropping `@voyant-travel/distribution` from the subset drops the three `channel-push` jobs. Every job POSTs `SCHEDULED_JOB_ROUTE` (`/__voyant/scheduled?cron=<expr>`).
- `getManagedProfileWorkflowManifest(project)` → the profile's workflow definitions at `{ id, config }` grain (voyant#2925), for active modules only. Sources the value from the module's own manifest const (`bulkReindexProductsWorkflowManifest`), with a drift test.
- `STANDARD_OPERATOR_SCHEDULED_JOBS` — the full all-modules set.

The `operator` starter now derives `OPERATOR_CRON_JOBS` **and** its cron dispatch constants from `STANDARD_OPERATOR_SCHEDULED_JOBS` instead of a hand-maintained list, appending only its deployment-local `external-cruise-catalog-refresh` (`@voyant-travel/cruises` is not a standard framework module). One source of truth for operator + managed; the emitted job set is byte-identical to before.

## Acceptance

Given only a valid `voyant.managed-profile.v1` snapshot and the installed framework version, Cloud can now compute the full Cloud Scheduler job set and the workflow release manifest with zero build involvement. ✅

## Verification

- `pnpm verify:fast` green; framework + operator typecheck/lint/build.
- 7 unit tests: full snapshot → 6 standard jobs + 1 workflow; subset `[bookings, catalog]` → `outbox-drain, draft-reaper, promotion-boundary-scheduler` (no `channel-push`, distribution inactive); `[bookings]` → no `draft-reaper` (catalog inactive); module tagging; workflow `{id,config}` grain + drift-vs-const.

## Scoped follow-up (out of this PR)

This exports the manifests for **provisioning** (the issue's acceptance). Executing those crons on a managed deployment is a separate slice: the managed runtime (`createManagedProfileApp`) currently mounts **no** `scheduled()` dispatcher — only the operator starter does — so a scheduler job POSTing `/__voyant/scheduled` on a managed image would no-op today. Worth a follow-up issue (a framework-owned managed scheduled dispatcher) so the provisioned jobs actually run.

### DIFF
--- a/.changeset/managed-profile-jobs-manifests.md
+++ b/.changeset/managed-profile-jobs-manifests.md
@@ -1,0 +1,31 @@
+---
+"@voyant-travel/framework": minor
+---
+
+Export the standard profile's scheduled-jobs and workflow manifests from the
+framework so Voyant Cloud can provision managed deployments with no build
+(voyant#3032).
+
+Managed-profile deployments run a fixed `voyant-operator-runtime:<framework-version>`
+image with no build step, so Cloud needs both the Cloud Scheduler job set and the
+workflow release manifest derivable purely from a profile snapshot. Both were
+reachable only from `starters/operator` (the cron list) or a build artifact (the
+workflow manifest).
+
+New subpath `@voyant-travel/framework/managed-jobs`:
+
+- `getManagedProfileScheduledJobs(project)` — the cron jobs to create for a
+  snapshot, each `{ id, cron, description, route, module }`, gated by the
+  resolved module subset (always-on framework infra like `outbox-drain` plus one
+  job per active owning module — e.g. dropping `@voyant-travel/distribution`
+  drops the `channel-push-*` jobs). Every job POSTs `SCHEDULED_JOB_ROUTE`
+  (`/__voyant/scheduled?cron=<expr>`).
+- `getManagedProfileWorkflowManifest(project)` — the profile's workflow
+  definitions at `{ id, config }` grain (voyant#2925), for active modules only.
+- `STANDARD_OPERATOR_SCHEDULED_JOBS` — the full all-modules set.
+
+The `operator` starter now derives `OPERATOR_CRON_JOBS` (and its cron dispatch
+constants) from `STANDARD_OPERATOR_SCHEDULED_JOBS` instead of a hand-maintained
+list, appending only its deployment-local `external-cruise-catalog-refresh`
+(`@voyant-travel/cruises` is not a standard framework module) — so the operator
+and a source-free managed deployment provision the same jobs from one source.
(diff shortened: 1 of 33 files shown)


mihaipxm:
Good catch — fixed in 06f6a36.

You're right that registering the workflow without `promotionAffectedAllFilter` would leave `promotions.reindex-all-products` un-triggered on managed deployments (the composition declares both `workflows` and `eventFilters` for promotions). Added a sibling export:

- `getManagedProfileEventFilters(project)` → the `event → workflow` routing bindings for active modules, keyed on the same resolved module set (mirrors `STANDARD_PROFILE_EVENT_FILTERS` ↔ `composition-lazy`'s `eventFilters`). Cloud registers these alongside `getManagedProfileWorkflowManifest(...)`.
- `ManagedEventFilterEntry { id, eventType, manifest? }` — the serializable descriptor (the runtime entry carries the full where/input/target manifest the event router needs).
- Tests assert the filter is emitted for active modules and that its `manifest.targetWorkflowId` pairs with the workflow id.

Also fixed the CI `Architecture checks` failure (regenerated composition dep-paths + package dist-configs for the new `@voyant-travel/framework/managed-jobs` subpath edge).

On the earlier order-parity note: the reordered `OPERATOR_CRON_JOBS` is intentional (framework infra first, then per-module) — ids/crons/descriptions are identical and both the Cloud Scheduler emitter and `entry.ts` dispatch key on id/cron, not list position, so no functional change.

PR #3258, opened 2026-07-11
refactor(operator): delete final SmartBill runtime bridge

mihaipxm:
## Summary
- upgrade `@voyant-travel/plugin-smartbill` to 0.140.0 and bind its typed `smartbill.runtime-host` port
- compose package-owned plugin modules generically and use package-owned subscribers and settlement pollers
- delete the 930-line Operator SmartBill runtime bridge and its 397-line test suite
- enforce zero package-ID runtime bridges in the standard Node starter checker

## Verification
- 17 framework runtime-composition tests
- 48 focused Operator composition, route, and SmartBill host tests
- SmartBill, runtime-port, standard-starter, Trips, and Flights architecture checks
- deployment graph and OpenAPI coverage checks
- dependency version, dep-path, dist-config, changed lint, and changed quality checks

Full package typecheck was started but stopped after prolonged resource starvation from concurrent repository typechecks; CI remains the full typecheck gate. The pre-push full-repository suite also encountered the existing stacked-base Relationships manifest failure; scoped checks above pass.

## OpenAPI
SmartBill 0.140.0 does not declare an OpenAPI document in its package manifest, so the single explicit coverage allowlist remains with an updated versioned reason.


mihaipxm:
Folded into consolidated governance PR #3256 so the SmartBill bridge deletion ships in the same reviewed wave.

PR #3259, opened 2026-07-11
feat(starter): extract minimal graph-native Node host

mihaipxm:
## Summary

- release the Operator starter as a literal minimal consumer project with only config, package metadata, seed entry, and optional convention folders
- move generic Node boot and static admin hosting into `@voyant-travel/operator-runtime`
- resolve the versioned standard BOM into the inspectable `.voyant` graph and preserve file-based API, admin, module, workflow, job, subscriber, and link discovery
- remove the synthetic managed-profile artifact from graph-native boot
- make `createVoyantApp` fail loudly when its legacy registry cannot compose selected standards, while managed boot explicitly composes selected graph and remaining selected legacy factories
- migrate framework OpenAPI generation to the selected graph and regenerate the previously missing standard package documents

## P1 resolution

`createVoyantApp` no longer filters graph-owned standards. Its implicit legacy-standard mode lists missing factories and throws. The generic Node host calls it with `standard: false` after composing the admitted graph. Framework OpenAPI resolves the same standard BOM graph, composes package-owned factories through conforming inspection ports, adds only selected factories that remain in `composition-lazy`, and also calls `createVoyantApp` with implicit standards disabled.

This means incomplete framework apps/specs can no longer be produced through silent filtering.

## Verification

- `pnpm verify:minimal-starter`
- `pnpm verify:standard-node-starter`
- Framework focused Vitest: create-app, project resolver, project artifacts, managed runtime, API/workflow conventions
- `pnpm --filter @voyant-travel/openapi generate`
- `pnpm --filter @voyant-travel/openapi test`
- `node scripts/generate-framework-bom.mjs`
- `pnpm --filter @voyant-travel/operator-runtime lint`
- `pnpm --filter @voyant-travel/operator-runtime test`
- `git diff --check`

The isolated acceptance test packages the release archive, installs it offline, emits the graph, verifies local convention discovery, and boots/probes the Node server.

The package-scoped TypeScript build traverses the full Framework BOM and exhausted Node's default 4 GB heap locally. Its build/typecheck scripts now use the repository's established 14 GB heap setting; CI remains authoritative for that compile lane. Full Turbo was intentionally not run.

## Remaining bridges

- `packages/framework/src/composition-lazy.ts`: 531 lines, with 15 module and 8 extension factories. Graph-native managed boot and OpenAPI now select these explicitly; deletion requires package-owned runtime factories for those 23 entries.
- `starters/operator/src/api/composition.ts`: only `@voyant-travel/plugin-smartbill` remains in `operatorGraphRuntimeBindings`; this stack deliberately does not modify it.
- `createVoyantApp` remains as a final Hono assembly API and a fail-loud legacy front door. It is no longer a source of standard selection authority for the generic host or framework OpenAPI.
- The existing internal `starters/operator` tree remains a development/deployment application. Release packaging no longer copies it; deleting that internal app requires the separate SmartBill cutover and remaining deployment-specific admin/client migration.


mihaipxm:
Folded into #3251 so the P1 createVoyantApp/OpenAPI fix lands in the base authority cutover.

PR #3263, opened 2026-07-12
Fix generated starter public API convention

mihaipxm:
## Summary
- align project route discovery with the generated `src/api/public` starter directory
- exercise a real public route through graph emission and the generated Hono adapter
- fix the generated `.env.example` so it declares `DATABASE_URL`
- update the active ADR and architecture docs

## Root cause
The packaged starter exposed `src/api/public`, while framework convention discovery scanned `src/api/store`. Existing acceptance only exercised an admin route, so the mismatch passed while consumer-authored public routes were silently omitted.

## Verification
- `pnpm --filter @voyant-travel/framework exec vitest run src/project-conventions.test.ts src/project-api-conventions.test.ts`
- `node --test scripts/tests/check-standard-node-starter.test.mjs scripts/tests/minimal-starter-acceptance.test.mjs`
- `pnpm --filter @voyant-travel/framework typecheck`
- focused Biome check and `git diff --check`

### DIFF
--- a/.changeset/clean-public-api-starter.md
+++ b/.changeset/clean-public-api-starter.md
@@ -1,0 +1,7 @@
+---
+"@voyant-travel/framework": patch
+---
+
+Discover application public API routes from the generated starter's
+`src/api/public` convention and validate that a new starter can build and run a
+route authored there.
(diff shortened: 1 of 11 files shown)


mihaipxm:
Superseded by #3282, which addresses this PR's reviewer findings and merged in cda53b6b8cc164f6378b18af10d85ec314a15209.

PR #3266, opened 2026-07-12
Remove dead Operator API facades

mihaipxm:
## Summary
- delete four unreferenced Operator compatibility facades whose behavior is package-owned
- make domain and storage authority checkers reject those files returning
- extend the standard starter gate with the same deletion ratchet

## Removed
- `src/api/lib/catalog-context.ts`
- `src/api/runtime/payment-config.ts`
- `src/api/runtime/booking-payment-policy-runtime.ts`
- `src/api/runtime/media-runtime.ts`

## Verification
- 18 focused checker tests passed
- domain, storage/media, and standard starter checkers passed
- `git diff --check`

Stacked on #3265; no published package behavior changes.

PR #3275, opened 2026-07-13
move custom field backfill into package migration

mihaipxm:
## Summary
- replace the starter-owned one-off custom-field script with an automatic post-cutline Relationships migration
- validate entity types, definitions, target columns, and orphan IDs before atomically copying values
- preserve current JSON values, refuse unexpected dependencies, and drop the legacy table only after successful validation

## Verification
- 23 focused Relationships tests and package typecheck
- Drizzle migration check and no-op generation
- migration boundary and frozen-cutline checks
- standard starter checker tests and checker
- Biome and `git diff --check`

Stacked on #3266. CI database replay is the final live SQL gate.


mihaipxm:
Superseded by #3282, which addresses this PR's reviewer findings and merged in cda53b6b8cc164f6378b18af10d85ec314a15209.

PR #3376, opened 2026-07-15
Unblock team-management release plan

mihaipxm:
Removes the ignored TypeScript config package from the mixed team-management changeset. Changesets rejects files that combine ignored and publishable packages.

Verification: `pnpm release:plan` succeeds and includes Framework 0.46.0 plus the pending product releases.


mihaipxm:
The current branch head has been recreated. GitHub records Mihai Ungureanu as both author and committer; no agent or model attribution is present in the commit metadata.

PR #3377, opened 2026-07-15
Version packages before publishing pending releases

mihaipxm:
## Summary
- prioritize Changesets release PR creation when releasable changesets and unpublished local packages coexist
- publish pending package versions only after their changesets have been applied and merged
- cover the release state machine with focused tests and document the sequencing

## Verification
- `pnpm verify:release-config`
- `pnpm exec biome check scripts/release-plan.cjs scripts/tests/release-plan.test.mjs package.json`
- `git diff --check`

This prevents new package placeholder versions from being published before the release PR applies their intended version bump.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -45,9 +45,9 @@ Handles package releases:
 
 Notes:
 
-- the push flow publishes packages whose local version is ahead of npm, then creates Changesets release PRs without building the package graph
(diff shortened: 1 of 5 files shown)


mihaipxm:
The current GitHub commit object for `a08b72cc2ff1d70ba4d2801ada87e79fa8dca313` records Mihai Ungureanu <mihai@pixelmakers.com> as both author and committer. There is no agent or model attribution in the branch metadata.

PR #4666, opened 2026-08-15
fix(catalog-react): classify a failed Commit by its own outcome

### LINKED ISSUE #4662: A failed Commit always shows the generic fallback: the recovery mapper ignores the commit_result envelope
## Summary

Every commit-time quote or hold failure shows the operator the generic fallback — *"The Booking Session rejected this request. Review the booking and try again."* — even though a precise, actionable message for that exact condition already exists and is already wired up. The recovery mapper only understands the `rejected` envelope, and a failed Commit does not use it.

## The defect

`packages/catalog-react/src/booking-engine/session-outcomes.ts`:

```ts
export function bookingSessionRecoveryV1(
  outcome: BookingSessionOutcomeV1,
): BookingSessionRecoveryV1 {
  if (outcome.kind !== "rejected") return "unknown"
  switch (outcome.error.kind) {
    …
    case "quote_expired":
    case "quote_superseded":
    case "quote_required":
      return "quoteChanged"
    …
  }
}
```

A rejected **hold** returns `{ kind: "rejected", error: { kind: "quote_superseded", … } }` and maps correctly to `quoteChanged`.

A failed **commit** returns a different envelope entirely:

```json
{"kind":"commit_result","outcome":{"kind":"quote_failure","nextAction":"request_fresh_quote","reason":"superseded"}}
```

`kind` is `commit_result`, so the guard on the first line returns `"unknown"` before the switch is ever reached. The `quote_failure` outcome — and its `reason` (`superseded` / `expired`) and its `nextAction` (`request_fresh_quote`) — are all discarded.

## What the operator loses

The correct string is already authored and translated:

> **quoteChanged** — "The price changed or expired. Review the refreshed total and try again."

Instead they get the `unknown` fallback, which names no cause and suggests no action. The server told the client precisely what went wrong and precisely what to do about it (`request_fresh_quote`); none of that reaches the person who could act on it.

Observed on a production deployment: an operator creating a booking hit this three times in a row and had no way to tell whether the problem was the product, the traveller data, permissions, or the price. Reading the actual cause required opening devtools and inspecting the commit response body.

## Asks

1. **Map the `commit_result` envelope.** A `quote_failure` should resolve to `quoteChanged` (or a dedicated message that distinguishes `superseded` from `expired`), and a `hold_failure` to `availabilityChanged`, consistent with how the `rejected` envelope is already treated.
2. **Consider acting on `nextAction` rather than only reporting it.** `request_fresh_quote` is a machine-readable instruction the form is capable of following; re-quoting and retrying once would resolve the common case without troubling the operator at all.
3. **Make `unknown` loud in development.** A silent fallback that swallows a fully-structured outcome is how this survived — the fallback should be reachable only for genuinely unrecognised shapes, not for a first-class result type.

## Note

Both `quote_failure` and `hold_failure` also appear as settlement-commit rejection reasons in the tenant event outbox on this deployment, so this envelope is not rare.

mihaipxm:
Closes #4662.

## The defect

`bookingSessionRecoveryV1` opened with `if (outcome.kind !== "rejected") return "unknown"`. A failed Commit does not use that envelope — it answers

```json
{"kind":"commit_result","outcome":{"kind":"quote_failure","nextAction":"request_fresh_quote","reason":"superseded"}}
```

so the guard returned before the switch and the operator got the `unknown` fallback for a fully-structured outcome whose precise message (`quoteChanged` — "The price changed or expired…") was already authored and translated.

## Ask 1 — map the `commit_result` envelope

`commitOutcomeRecovery` classifies the Commit union, with `idempotent_replay` unwrapped to the outcome it replays:

| Commit outcome | recovery |
|---|---|
| `quote_failure` | `quoteChanged` |
| `hold_failure` | `availabilityChanged` |
| `revision_mismatch` | `revisionConflict` |
| `supplier_pending`, `supplier_in_doubt`, `component_commit_pending` | `commitInFlight` *(new)* |
| `supplier_failed` | `supplierUnavailable` *(new)* |
| `proposal_acceptance_required` | `proposalAcceptanceRequired` *(new)* |
| `committed`, `component_bookings_committed`, `payment_required` | `unknown` — successes, nothing to recover from |

The three new classes exist because no existing one was true. `commitInFlight` is the one that matters beyond wording: telling an operator a supplier-pending Commit was *rejected* invites the resubmit that books twice. On the rejection side the same reasoning moves `payment_in_flight`, `supplier_operation_active` and `commit_already_consumed` off `unknown` — re-quoting a Session with a payment in flight is voyant#4636.

en + ro strings added for all three.

## Ask 3 — make `unknown` loud

Both switches are now exhaustive over their contract unions, so **a kind added to either contract fails the build here** rather than silently joining the fallback. The remaining `default` is reachable only for a shape this build does not know (a newer server) and `console.warn`s outside production.

Kinds that genuinely have no remedy to name — `session_expired`, `invalid_selection`, `renewal_not_allowed`, … — are listed explicitly and return `unknown` silently, so the warning stays worth reading.

## Ask 2 — act on `nextAction`, without gambling with the commit key

`nextAction` and the unwrapped `commitOutcome` are now on `BookingSessionJourneyError`, and `bookingSessionContinuationIsStale()` says whether a stored continuation is dead.

I did **not** auto re-quote and retry. Two reasons: the message the operator would skip says "review the refreshed total", and committing at a price they never saw is a money decision that is not ours to make; and a silent retry under a live continuation is how you mint a second booking.

What the form does instead explains the report's "three times in a row". `attemptRef` was cleared only on success, so after a `quote_failure` an unchanged resubmit replayed the *same superseded* `quoteId` and failed identically, forever. It is now dropped when — and only when — the outcome says the Quote, Hold or revision it pins is gone (`request_fresh_quote`, `request_new_hold`, `request_hold_for_expected_quantity`, `refresh_session_state`). The next submit runs a fresh Create → Quote → Hold → Commit under a new idempotency key, which is safe precisely because those outcomes mean no booking was created. It is kept for `return_idempotent_result`, `await_payment_outcome` and the supplier actions, where it is the only key that can retrieve the original result.

## Tests

New `session-outcomes.test.ts` (15 tests) plus a journey-level test driving Create → Quote → Hold → a `quote_failure` Commit through the real helper and asserting `quoteChanged` at the catch block. One test iterates `bookingLifecycleCommitOutcomeKindV1.options` and asserts every declared Commit kind classifies without warning — so the contract, not a hand-written list, is what this is checked against.

`bookingSessionContinuationIsStale` is unit-tested in `catalog-react`; the form's one-line use of it is not, as `manual-booking-create-form.tsx` has no test harness today.

## Noted, not fixed

`commitBookingSessionJourneyV1` throws on `component_bookings_committed` — a success it cannot express through a single-`bookingId` result. Pre-existing, out of scope here, and now at least classified rather than reported as a rejection.

## Verification

- `catalog-react` booking-engine suite: 36 passed
- `tsc --noEmit` on `catalog-react` and `bookings-react` typecheck projects: 0 errors
- `biome check` on both packages: clean
- `i18n:check:packages`: passed, 49 definition sets

Both packages are private, so no changeset.

### DIFF
--- a/packages/bookings-react/src/components/manual-booking-create-form.tsx
+++ b/packages/bookings-react/src/components/manual-booking-create-form.tsx
@@ -25,6 +25,7 @@ import {
 import {
   type BookingSessionJourneyContinuation,
   BookingSessionJourneyError,
+  bookingSessionContinuationIsStale,
   commitBookingSessionJourneyV1,
   createBookingJourneyApi,
   useOfferPreview,
@@ -1853,6 +1854,18 @@ export function ManualBookingCreateForm({
       // The typed outcome carried the list; keep it. Collapsing it back into
       // the sentence above is what made the server's enforcement invisible.
       setUnsatisfied(cause instanceof BookingSessionJourneyError ? (cause.unsatisfied ?? []) : [])
+      if (
+        cause instanceof BookingSessionJourneyError &&
+        bookingSessionContinuationIsStale(cause.outcome)
+      ) {
+        // The server said the Quote, Hold or revision we are holding is dead.
+        // Keeping the attempt would replay the same dead continuation on every
+        // resubmit, which is how one operator hit the same failure three times
+        // in a row (voyant#4662). Dropping it makes the next submit run a fresh
+        // Create → Quote → Hold → Commit under a new idempotency key — safe
+        // precisely because these outcomes mean no booking was created.
+        attemptRef.current = null
+      }
     } finally {
       submissionRef.current = false
       setSubmitting(false)
--- a/packages/bookings-react/src/i18n/en-create-list.ts
+++ b/packages/bookings-react/src/i18n/en-create-list.ts
@@ -68,6 +68,8 @@ export const bookingsUiEnCreateList = {
         commitRejected: "The booking could not be committed from the current selection.",
         commitInFlight:
           "This booking is still being confirmed. Do not submit it again — refresh in a moment to see the result.",
+        alreadyCommitted:
+          "This booking session was already completed by another attempt. Find the booking it created instead of submitting again.",
         supplierUnavailable:
           "The supplier could not confirm this booking. Choose different inventory or send it for manual review.",
         proposalAcceptanceRequired:
--- a/packages/bookings-react/src/i18n/en-create-list.ts
+++ b/packages/bookings-react/src/i18n/en-create-list.ts
@@ -66,6 +66,12 @@ export const bookingsUiEnCreateList = {
         requirementsChanged:
           "The booking details this product requires changed. Review the refreshed form and try again.",
         commitRejected: "The booking could not be committed from the current selection.",
+        commitInFlight:
+          "This booking is still being confirmed. Do not submit it again — refresh in a moment to see the result.",
+        supplierUnavailable:
+          "The supplier could not confirm this booking. Choose different inventory or send it for manual review.",
+        proposalAcceptanceRequired:
+          "The proposal changed since it was accepted. Have the traveller accept the current version and try again.",
         notAuthorized: "You do not have permission to commit this booking.",
         unknown: "The Booking Session rejected this request. Review the booking and try again.",
       },
--- a/packages/bookings-react/src/i18n/messages-create-list.ts
+++ b/packages/bookings-react/src/i18n/messages-create-list.ts
@@ -54,6 +54,9 @@ export type BookingsUiCreateListMessages = {
         selectionIncomplete: string
         requirementsChanged: string
         commitRejected: string
+        commitInFlight: string
+        supplierUnavailable: string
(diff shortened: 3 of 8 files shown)


mihaipxm:
Good catch — verified at `sessions-service.ts:1377-1416`: `commit_already_consumed` is only reached after `getCommitByIdempotency` misses, so nothing is retrievable under that caller's continuation and "refresh in a moment" is a lie.

Split out as its own `alreadyCommitted` class in 7287d1dd97, pointing the operator at the booking that already exists rather than telling them to wait. `payment_in_flight` and `supplier_operation_active` stay on `commitInFlight`, where the work genuinely is still moving.

The continuation is deliberately still kept for this outcome: it is useless, but discarding it would let the next submit run a fresh Create → Quote → Hold → Commit and mint a second booking for the same intent.

PR #4667, opened 2026-08-15
Issue the invoice a booking creates, and stop billing a buyer the invoice cannot name

### LINKED ISSUE #4653: Invoices created from a booking are never issued, so external allocation, accounting sync and the invoice attachment never happen
## Summary

An invoice created from a booking is never **issued**. `createBookingFromCommand` renders a PDF and stops:

```ts
let invoiceDocument: BookingCreateResult["invoiceDocument"] = { status: "not_requested" }
if (documentGeneration.invoiceDocument) {
  const requested = await financeService.renderInvoice(tx, invoice.id, { format: "pdf" })
  ...
}
```

Nothing calls `issueInvoiceFromBookingCommand`. For a deployment whose invoice number series has an `externalProvider`, that has three consequences, and together they mean a paying customer never receives an invoice.

## 1. The number is never allocated

`resolveInvoiceNumber` deliberately mints a placeholder for an external series and defers the real number to issuance:

```ts
if (series.externalProvider) {
  return {
    invoiceNumber: pendingExternalInvoiceNumber(scope),
    seriesId: series.id,
    sequence: null,
    status: "pending_external_allocation",
  }
}
```

That contract is correct. Nothing ever completes it, so the invoice sits at `PENDING-INVOICE-<uuid>` / `pending_external_allocation` indefinitely.

## 2. The accounting app is never told

`invoice.issued` carries the whole external-allocation contract — `externalAllocationRequired`, `externalProvider`, `externalConfigKey`, `externalSeriesId`, `externalPlaceholderNumber` — and is declared as an outbound webhook. An installed accounting app subscribes to it.

Because issuance never runs, the event never fires. On a production deployment with the app installed and **active**, all five of its subscriptions read `last_delivery_at: null` — `invoice.issued`, `invoice.payment.recorded`, `invoice.proforma.converted`, `invoice.voided`, `invoice.payment.recorded`. The integration has never received a single event in its lifetime.

The only way to get a real number today is an operator opening the invoice and clicking the app's own **Send to SmartBill** panel button, per invoice, by hand.

## 3. The confirmation email goes out with no invoice attached

`resolveBookingDocumentBundle` collects contract attachments and invoice renditions, but only renditions that are `ready`:

```ts
.where(and(eq(invoiceRenditions.invoiceId, ...), eq(invoiceRenditions.status, "ready")))
```

With no issuance there is no ready rendition, so the bundle has nothing and the booking confirmation is sent without the invoice. This is exactly what a customer reported: *"nu am mai primit nimic pe e-mail, cum ar fi factura sau contract, doar receipt … fara sa existe un fisier pdf downloadabil."*

Worth noting the retry machinery for this **already exists and is correct** — `subscriber-runtime.ts` retries bundle delivery on `invoice_rendered` and `contract_document_generated`. It simply never has anything to deliver, because step 1 never happens.

## Asks

1. **Issue the invoice when it is created for a booking**, so `invoice.issued` fires and external allocation completes without an operator clicking anything. The natural trigger is the same condition that already creates it — `documentGeneration.invoiceDocument`, or a paid schedule (`shouldCreateInvoice`).

2. **Keep it in sync afterwards.** The subscriptions for `invoice.payment.recorded`, `invoice.voided` and `invoice.proforma.converted` exist and are active; confirm they actually emit once issuance is real, so a payment or a void reaches the accounting system rather than only the initial document.

3. **Do not send a notification whose template declares a document attachment until that attachment is resolvable.** The bundle path already returns `no_attachments`, and the retry-on-render path already exists — the booking confirmation should defer rather than send an email that promises an invoice and carries none.

## Related

- The buyer's fiscal code never reaches the payload either: `clientVatCode` is hardcoded `null` in `buildInvoiceIssuedPayload` despite `bookings.contact_tax_id` being on the row. PR follows separately.
- The admin manual-booking form collects no billing address, so even a successfully issued invoice can be fiscally invalid. Filed separately.

### LINKED ISSUE #4654: The manual booking form collects no billing address, so invoices issued from it are fiscally invalid
## Summary

The admin **New booking** form never collects a billing address, so a booking created by an operator carries `contact_address_line1`, `contact_city`, `contact_region`, `contact_postal_code`, `contact_country` and `contact_tax_id` all `null`. A fiscal invoice issued from such a booking has no buyer address, which makes it invalid under Romanian invoicing rules (art. 319 Cod Fiscal requires the buyer's name **and address**).

Observed on a production deployment: an invoice issued to a real customer rendered as

```
Client: <name>
CIF:
Adresa: -, -
Judet: -
```

## Why the storefront is fine and the admin form is not

The storefront checkout already captures the whole billing block, and there is an explicit comment saying so:

```ts
// The whole declared address, not just the country. The tracer
// (voyant#4039) projected the rest away, so a caller that filled the
// billing step lost every line of it at the Session edge and the
// Booking's contact_* columns came back empty. `region` is new
// (voyant#4290); the other four were declared all along and simply
// never survived to the commit.
address: pruneEmpty({
  line1: …, line2: …, city: …, region: …, postal: …, country: …,
}),
```

So `bookings.contact_*` is populated on the storefront path. The manual admin path only selects a **Person** — and a person record's address is not copied onto the booking's contact columns — so those columns stay empty and nothing warns the operator.

`buildManualBookingContactInput({ billTo, contact })` already exists on the submit path, and `createBookingFromCommand` accepts `contactAddressLine1` and friends. The form simply never gathers them.

## Impact

Every invoice issued from an operator-created booking is missing mandatory fiscal fields. The operator has no indication anything is wrong: the form does not ask, the booking detail does not flag it, and issuance succeeds.

## Asks

1. **Collect billing details in the manual booking form** when an invoice will be issued — address line, city, region, postal code, country, and tax id for a business buyer. Prefilling from the selected person's address (and letting the operator correct it) would match how the storefront billing step behaves.

2. **Validate before issuing rather than after.** A booking with no buyer address should not be able to produce a fiscal invoice silently. Failing at issuance is better than issuing an invalid document, and failing at booking creation with a clear prompt is better still.

3. Consider surfacing the gap on the booking detail — an invoice-bearing booking with no billing address is actionable information the operator currently has no way to see.

## Related

- `clientVatCode` is hardcoded `null` in the invoice payload even when `bookings.contact_tax_id` is populated, so the fiscal code cannot reach an accounting integration regardless of what is collected here. PR follows separately.
- Invoices created from a booking are never issued at all, filed separately.

mihaipxm:
Closes #4653. Closes #4654.

## The chain that was broken

A booking created with "Generate invoice" produced an invoice row and stopped there. Four things followed from that, and together they are why a paying customer never received one.

| | before | after |
|---|---|---|
| the invoice is issued | never — nothing called the issuing path | booking create issues it |
| `invoice.issued` reaches an installed app | never; all five subscriptions read `last_delivery_at: null` on a live deployment | written to the outbox with the booking |
| an external series gets a real number | never; stuck on `PENDING-INVOICE-<uuid>` unless an operator pressed the app's own button per invoice | the event carries the allocation contract, so the app allocates |
| `invoice.payment.recorded` for an already-paid schedule | never — `createPayment` was called with no runtime at all | written to the same outbox |
| the confirmation email | sent immediately, promising an invoice and carrying none, recorded as delivered so the retry could never repair it | deferred until the document resolves, then re-delivered |
| the buyer's address on the invoice | `Adresa: -, -` / `Judet: -` for every operator-created booking | collected in the form, required when a document will be issued |
| `clientVatCode` | hardcoded `null` while `bookings.contact_tax_id` sat on the row | populated for a company buyer |

## Why the outbox and not the event bus

`createBookingMutation` runs inside a leased transaction, and `eventBus.emit` is the wrong instrument there on three independent grounds:

- **It escapes the transaction.** With `outbox: true` the durable bus captures on `c.get("db")` — the per-request pooled client, not the transaction handle. A rollback after the emit leaves a live `invoice.issued` row that the drain delivers to every installed app, for an invoice that does not exist.
- **Subscribers race the open transaction.** They run on other connections under read-committed, so they cannot see the booking they were told about. On the 1-connection pool the integration tests use, the capture insert queues behind the transaction that owns the only socket and self-deadlocks.
- **It is absent on two of the four call paths.** `commitPaidSession` — the paid-checkout settlement driven by the `payment.completed` subscriber — passes no `financeRuntime`, so `runtime.eventBus?.emit(...)` is a silent no-op there.

`insertOutboxEvents(tx, …)` is the established pattern (11 packages use it; `booking.created` and `booking.confirmed` already go this way from this exact function). So `createBookingMutation` hands its events up on the outcome and `insertBookingCreatedOutbox` writes all of them in one atomic statement. `FinanceServiceRuntime.domainEventSink` is the seam that lets `createPayment` participate without knowing where it is running.

Because `invoice.issued` is `visibility: "external"`, an outbox row reaches installed apps through exactly the same chain as a live emit: drain → `bus.deliver` → the `graph-app-webhooks` subscriber → `infra_webhook_deliveries` → the worker POST. No new wiring.

## Not issuing an invalid invoice is not a regression

#4654 asks that a booking with no buyer address should not silently produce a fiscal invoice. Refusing to issue one is free here: issuance is new behaviour, so an invoice left as a draft is exactly where every booking-created invoice already sat. Nothing loses behaviour it had, and the operator is told which fields are missing instead of finding out from an accounting system.

`missingFiscalBillingFields` in `@voyant-travel/bookings-contracts` is one rule read by three surfaces — the form (blocks submit), booking create (leaves the invoice a draft), and the booking detail (names the gap). It is the invoicing floor, not a jurisdiction rule: a buyer's name and address, plus a fiscal code when the buyer is a company. Romania's art. 319 is the case that surfaced it; every regime wants the same minimum.

## The confirmation email

The readiness gate was `payment_complete`-only, so a `booking_confirmed` template declaring an invoice attachment sent the moment the booking was confirmed — before the invoice was rendered and before legal generated the contract, both asynchronous off the same event. The gate is now the **template's own declaration**, not the target type, so a confirmation that promises nothing still sends immediately. `getBookingEventDocumentContext` now returns `unresolvedDocumentTypes`, so an unresolvable brochure cannot hold back an email whose invoice is ready. The `invoice.rendered` / `contract.document.generated` subscribers re-drive `booking_confirmed` as well as `payment_complete` — the latter needs a paid session and a paid-in-full booking, which a confirmation does not.

The stored sentinel keeps its `payment_document_bundle_not_ready:` value even though it now covers more than payments: rows carrying it exist in live deployments and the retry predicate matches on it, so renaming it would strand every run already waiting. It is a named constant with the reason on it.

## Scope note — the invoice PDF

Fixing issuance closes the reported loop for a deployment with an accounting app: `invoice.issued` → the app allocates a number and PUTs the finished PDF to `/v1/app/finance/documents/:id/artifacts/provider-pdf` → `bindInvoiceRendition` → `invoice.rendered` → the retry re-delivers the confirmation with the invoice attached.

A deployment **without** such an app still cannot produce an invoice PDF at all, and that is a separate, pre-existing gap: `renderInvoice` inserts a `pending` rendition that nothing in the repo ever fulfils, and `POST /invoices/{id}/generate-document` returns 501 because no deployment supplies `invoiceDocumentGenerator`. Finance is missing the four pieces legal has for contracts — a `document-renderer` resource on its manifest, a conformance-tested provider port, the already-written `createPdfInvoiceDocumentGenerator` wired into `createFinanceRuntime`, and a job or subscriber to drive it. Out of scope here; filed as #4668.

## Verification

- `pnpm verify:architecture` — passes (80 chain links).
- `biome check .` — 6910 files, 0 errors.
- `typecheck` — clean across finance, notifications, bookings-react, bookings-contracts.
- `bookings-react` 37 files / 245 tests, `bookings-contracts` 9 tests, `notifications` unit 23 files / 172 tests — all pass.
- `finance` `booking-create.test.ts` — 100 pass / 1 skipped, including 7 new integration tests driven through `executeFinanceStaffBookingCreateCommand` rather than the mutation, because the point is where the event lands.

Two pre-existing failure sets were confirmed identical on `main` at `18f623fd0a` against the same database, and are not from this branch:

- **finance**: 155 failures across 8 integration files (`currency-amount-checks`, `document-routes`, `payment-adapter-initiation`, `public-routes`, `routes`, `settlement-routes`, `supplier-invoices`, `unsynced-proforma-command`) — byte-identical counts on both branches. `currency-amount-checks` inserts `bookings` without `status`, which the schema declares `notNull()` with no default.
- **notifications integration**: the files interfere through the shared database (`TRUNCATE … deadlock detected`, cross-file FK violations); `main` fails the same suite, and the failing set shifts run to run on `main` too. Two untouched files run together are enough to reproduce it. Related to #4251.

While adding coverage I found `invoice_number_series` missing from `booking-create.test.ts`'s truncation list, so the one test that seeded an external series was silently deciding how every later test in the file numbered its invoices. Added.


mihaipxm:
Both P1s were real; fixed in 671479f and the merge that follows.

**Company buyer judged by `contactPartyType` alone.** Confirmed: the column is independently optional, `personId`/`organizationId` are mutually exclusive (`validateExclusiveBillingParty`), and nothing anywhere derives one from the other — so `create_booking`, `book_product` (optional `billingContact.partyType`) and the storefront can all produce an organization booking with a null party type. `isCompanyFiscalBuyer` now reads either signal. Covered by a contracts unit test and an integration test that books to `org_booking_create` with no party type and asserts the invoice stays a draft with no `invoice.issued`.

**Rendering an invoice whose issuance was blocked.** Also real, and worse than described in one way: it contradicts the decision on its own terms — we refused to make the document a fiscal record and then asked for its PDF. Fixed at both ends, because skipping the render alone leaves `POST /invoices/{id}/render` as a way back in:

- booking create requests no rendition for a blocked invoice, and reports `invoiceDocument: "not_available"` rather than `"not_requested"`, so "wanted and withheld" is distinguishable from "nobody asked";
- `listFinanceBookingDocuments` excludes `draft` outright. That is the durable guarantee — a draft is not a document anyone may be sent, whatever route gave it a rendition. `pending_external_allocation` deliberately stays in: it is issued and merely waiting on its number, and `applyExternalInvoiceNumber` moves it to `issued` in the same write that supplies one, so excluding it would strand legitimately-ready documents.

Both are pinned by tests, including one asserting a `draft` invoice with a `ready` rendition still leaves the confirmation deferred.

One correction on the `clientVatCode` half: #4664 landed the same field on main while this was in review, sending `contact_tax_id` unconditionally on the grounds that downstream e-invoicing derives taxpayer status from its presence. That is the better rule for the payload and supersedes the company-only gating I had written, so the merge takes main’s line. `isCompanyFiscalBuyer` stays where it belongs — the issuance gate — because "must this buyer have a fiscal code before we may issue" and "does this buyer have one" are different questions.

PR #4670, opened 2026-08-15
feat(bookings): record documents issued outside Voyant against a booking

### LINKED ISSUE #4657: No Tool can attach an already-issued contract or invoice to a booking, only generate a new one
## Summary

An operator can attach an already-issued document to a booking through the admin UI, but there is no Tool that lets the agent do it. An operator migrating historical bookings — or recording a booking whose contract and invoice were issued outside the system — cannot complete the job conversationally, because the documents can never be attached.

## What exists

Legal exposes exactly two contract-document Tools, and both **generate** from a template:

```ts
export type LegalContractDocumentPolicyMode = "generate" | "regenerate"
…
toolName: `${mode}_booking_contract_document`,
```

- `generate_booking_contract_document`
- `regenerate_booking_contract_document`

There is no third mode for a document that already exists on paper and simply needs recording against the booking.

Searching every `canonicalName` in `packages/` returns **no upload-style Tool at all** — nothing for booking documents, invoice attachments, or media. The closest match by name, `attach_departure_fleet_resource`, is unrelated (it links a fleet resource to a departure).

Meanwhile the admin UI has an **Upload document** control on a booking's Documents tab, and an invoice carries an Attachments tab, so the underlying capability exists — it is only missing from the Tool surface.

## Why it matters

The realistic migration path onto the platform is an operator entering historical bookings whose paperwork was issued elsewhere. That is exactly the case where delegating to the agent is most valuable: many bookings, each mechanical. Today the agent can create the person, and (once the reachability defect is resolved) create the booking, but the externally-issued contract and invoice have to be attached by hand afterwards, one booking at a time, which removes most of the benefit.

There is a fiscal dimension too. When an invoice was issued by an external accounting system, the authoritative document is that PDF. Being able to record it against the booking is not a convenience — it is how the booking becomes auditable.

## Asks

1. **A Tool to attach an already-issued document to a booking**, covering at least the contract and the invoice, recording the external document's own identity (series, number, issue date) rather than allocating a new one.
2. Make it explicit in the Tool surface that this is *recording*, not *issuing* — an external document must not consume a number from an internal series, and must not be re-rendered from a template.
3. If a general booking-attachment Tool is the better shape, that would cover the same need plus traveller documents (passport, visa, insurance), which the UI already supports and the Tool surface likewise does not.

## Related

The agent currently cannot reach booking-creation or payment-recording Tools either, which blocks the same workflow one step earlier — filed separately.

mihaipxm:
Closes #4657.

## What this adds

An operator migrating historical bookings holds paperwork that was issued somewhere else: a contract signed on paper, an invoice produced by the accounting system. The admin UI could attach those; no Tool could, so the agent could create the person and the booking and then had to hand the documents back to a human, one booking at a time.

- **`record_booking_document`** — records a contract, invoice, proforma, credit note, or traveller document against a booking.
- **`list_booking_documents`** — reads the collection so the agent can see what is already on file.

## Recording is not issuing

Ask 2 in the issue is the load-bearing one, and the shape enforces it rather than the prose:

- nothing here allocates a number from `invoice_number_series` or a contract series — the document keeps the number its own issuer gave it, in `issuedBy` / `issuedSeries` / `issuedNumber` / `issuedAt`;
- nothing renders from a template, so the recorded file stays the authority;
- no `invoices` row and no legal `contracts` row is created, so nothing here enters accounts receivable, settlement, or the contract lifecycle.

The Tool description says all three, and points at `generate_booking_contract_document` / `issue_invoice_from_booking` for the case where the operator does want Voyant to issue.

## Why a booking-attachment Tool rather than a third legal mode

Issue ask 3 offered this as the better shape and I think it is, for a reason beyond covering traveller documents. Legal's `generate`/`regenerate` modes operate on an *existing* legal contract and render its template; a contract issued on paper has no contract row, and minting one would mean fabricating a template version and a rendered body. Symmetrically, creating a finance `invoices` row for an externally-issued invoice pulls it into AR and settlement, which is not what "record the PDF" means.

`booking_documents` is already the booking's document shelf and already backs the **Upload document** control the issue cites — this widens it rather than inventing a surface.

The four commercial kinds are `contract`, `invoice`, `proforma`, `credit_note` — **exactly the words the customer portal's unified document view already uses** (`packages/storefront/src/customer-portal/validation-public/common.ts`), which maps `booking_documents.type` straight through its own enum. Picking any other word (I started with `receipt`) would have made the portal's booking detail fail to parse for that customer.

## Enforcement

- `ck_booking_documents_issued_identity` — a contract/invoice/proforma/credit-note row must carry the issuer's number and date. In the database, so no writer bypasses it; mirrored in the Zod schema so a caller is told which field is missing instead of being handed a constraint violation.
- `uq_booking_documents_issued_identity` — partial unique index on `(booking_id, type, coalesce(issued_series, ''), issued_number)`. Recording the same document twice replays the first record (`replayed: true`) instead of doubling it in the booking's audit trail; `onConflictDoNothing` plus a re-read settles the concurrent case.
- A `booking.document.record` action-ledger entry, written by both the Tool and the admin route so an agent recording and a person recording read the same way, and only for a recording that actually happened.
- Graph actions `bookings:documents:read` (sensitive-read — the collection holds passports) and `bookings:documents:record`.

## Verification

- Both migrations **executed against Postgres**, not reviewed: fresh database, full bookings migration chain in order, then inserts proving the check rejects an invoice with no number, the unique index rejects the duplicate, a differing series is a different document, and traveller documents still need nothing and repeat freely.
- `pnpm verify:architecture` green end to end. It caught two real gaps on the way: `list_booking_documents` was declared `risk: low` against a `sensitive` tier, and `record_booking_document` had no graph action.
- `verify:agent-write-coverage` baseline for bookings lowered 16 → 15; `booking/document` was the uncovered resource.
- Unit tests green in `bookings`, `bookings-contracts`, `bookings-react`, `storefront`; `biome check .` clean; `i18n:check` clean (en + ro).
- Builds green for every dependent of the three changed packages, including `apps/operator`.

**Not verified locally:** the four integration cases I added to `routes.integration-suite.ts`. That whole suite is red in this environment for unrelated reasons — 55/58 fail on `origin/main` too, every one with the same `SyntaxError` out of `res.json()`. The suite does not run in CI either (#4251).

## Notes for the reviewer

- `packages/bookings/openapi/admin/bookings.json` says "Do not edit by hand" but has no generator. I wrote one and it emitted a 5,252-line diff against the committed file, so the checked-in document has drifted a long way from its routes. I reverted the generator and hand-patched the document to stay in scope; that drift is worth its own issue.
- No `delete_booking_document` Tool. The issue does not ask for one and a destructive Tool wants its own approval design; the admin UI still deletes.


mihaipxm:
All five review points were real. Addressed in `9b88746`.

**P1 — replay identity was the number, not the document.** Correct, and the failure mode was the bad one: a distinct document silently returned as a replay with 201 and its file never stored. The unique index and the lookup that mirrors it now key on the whole issued identity — `(booking_id, type, coalesce(issued_by,''), coalesce(issued_series,''), issued_number, coalesce(issued_at,'-infinity'))`. Verified against Postgres: the same identity is rejected as a duplicate; a different issuer and the same number in another year are both stored as distinct documents.

**P1 — idempotent retries through the MCP action gate.** Confirmed by reading `tool-action-policy.ts`: `commandTargetField` makes this a server-owned generic target, so `executionKey` is `mcp-request:<fingerprint of the whole command>`, the preflight replays, and `duplicateDispatchError` throws `AUTHORIZATION_DENIED` before the handler runs. `record_booking_document` now owns its action policy (`actionPolicyEnforcement: "handler"` + `resolvesIdempotencyKeyServerSide`), so the unique index is the single replay authority and an identical retry returns `replayed: true`. The key stays off the wire, so no token crosses calls.

**P2 — validate `issuedAt` before constructing a Date.** Fixed; it is now the same ISO date-or-timestamp union the Tool schema uses, shared from `bookings-contracts`. `expiresAt` has the same latent shape but predates this change and is left alone rather than tightened on callers here.

**P2 — commit the document and its audit entry atomically.** Fixed, and the argument is exactly right: because the recording replays a document already on the booking, a failed append left a recording that no retry could ever audit. Both surfaces now go through one `recordBookingDocument` in `document-recording.ts` that runs the insert and the ledger append in a single transaction.

**P2 — ledger the admin document-list read.** Correct: the graph declared a required-ledger sensitive read that the route never wrote. `GET /{id}/documents` now appends `booking.document.read` with the capability identity from the declaration.

Re-verified after the fixes: `verify:architecture` green, migrations re-executed against Postgres, `bookings` / `bookings-contracts` / `bookings-react` / `storefront` unit tests green, `biome check .` clean, and every dependent of the changed packages builds.

PR #4687, opened 2026-08-15
Produce invoice documents without an accounting app

### LINKED ISSUE #4668: Nothing fulfils a requested invoice rendition, so a deployment without an accounting app can never produce an invoice PDF
## Summary

`financeService.renderInvoice` inserts an `invoiceRenditions` row with `status: "pending"` and its own docblock says what is meant to happen next:

```ts
/**
 * Request an invoice rendition. Creates a `pending` rendition row pointing
 * to a template; the actual rendering (HTML→PDF) is expected to be
 * performed out-of-band by a background job that updates the rendition to
 * `ready` with `storageKey` set.
 */
```

That background job does not exist. **Nothing in the repository reads `invoiceRenditions` where `status = 'pending'`** — no job, cron, queue consumer, worker or subscriber. `bindInvoiceRendition` always inserts a *separate* `ready` row and never touches the pending one, so a requested rendition is orphaned, not even marked `stale`.

So a Voyant deployment cannot produce an invoice PDF on its own.

## The three entry points and where each dies

- `POST /invoices/{id}/render` (`routes-invoice-documents.ts`) → a `pending` row nobody fulfils. With `?wait=true`, `waitForInvoiceRendition` polls for a terminal row for its full 30s and returns 202, because the row it is waiting on can never change.
- Booking create with `documentGeneration.invoiceDocument` (`service-booking-create.ts`) → the same orphan, returned to the caller as `{ status: "requested" }`.
- `POST /invoices/{id}/generate-document` (`routes-documents.ts`) → **HTTP 501**, `"Invoice document generator is not configured"`.

The 501 is the interesting one. `createStorageBackedInvoiceDocumentGenerator` and `createPdfInvoiceDocumentGenerator` are both written, exported, and tested — and their only consumers repo-wide are `scripts/verify-package-exports.ts` and a unit test. `createFinanceRuntime` never sets `invoiceDocumentGenerator` or `resolveInvoiceDocumentGenerator`, and no deployment supplies finance host options, so the field is `undefined` everywhere. The only place a generator is ever provided is `packages/finance/tests/integration/document-routes.test.ts`.

The one path that does yield a `ready` rendition today is a third party doing the rendering: an installed app PUTs finished bytes to `/v1/app/finance/documents/:id/artifacts/provider-pdf`. That is why a SmartBill-style deployment gets PDFs and a plain one gets none.

## Why it stayed invisible

`invoiceDocumentGenerator` is a plain optional field on `FinanceDocumentRouteOptions` — not a graph port. No `definePort`, no `providePort`/`requirePort`, no `selection.role`, no conformance test, and `packages/finance/src/voyant.ts` declares **no `resources` at all**, so finance never asks the deployment for a `document-renderer` or `document-storage`. Nothing at graph-check time can notice it is unfilled; it degrades to a runtime 501.

Contracts work because legal has all four pieces finance is missing:

| | legal (contracts) | finance (invoices) |
|---|---|---|
| resource declared | `#resource.document-renderer` + `#resource.document-storage`, both `required: true` | none |
| provider port | `legalDocumentArtifactProviderPort`, conformance-tested, selected as `legalDocumentArtifact: "standard"` in `operator-standard` | none — an optional options field |
| trigger | `booking.confirmed` subscriber, renders in-process | none |
| recovery | `legal.contract-document-operations` job, `every: "1m"`, `wakeup: true` | none — finance declares no jobs |

Legal's subscriber also carries the voyant#4634 fix directly above the relevant line: it used to `return` silently when no renderer was present, and now records the miss with `reason: "document_renderer_unavailable"`. Finance still has the un-fixed version of that behaviour and does not even record it.

## Asks

1. **Declare the resources.** `#resource.document-renderer` and `#resource.document-storage` on finance's manifest, so a deployment that cannot render says so at graph-resolution time rather than at 501.
2. **Make the generator a provider port**, with a conformance check, and select it in `operator-standard` — mirroring `legalDocumentArtifact`. `createPdfInvoiceDocumentGenerator` already exists; it needs wiring into `createFinanceRuntime`, not writing.
3. **Give it a trigger and a recovery leg.** Legal renders in-process off the event and keeps a 1-minute job for the tail. Whatever shape finance picks, the `pending` row `renderInvoice` writes must either be fulfilled or stop being written — an orphan row that reads as "requested" is worse than no row, because `?wait=true` and the notification bundle both treat it as work in flight.
4. **Record the miss.** When no renderer is available, say so on the invoice the way `recordUnfulfilledBookingContract` does, instead of leaving a silent orphan.

## Related

- voyant#4653 fixed issuance, which closes the reported loop **only** for a deployment with an accounting app: `invoice.issued` → the app allocates a number and PUTs the PDF back → `bindInvoiceRendition` → `invoice.rendered` → the notification retry delivers the confirmation with the invoice attached. A deployment without such an app still gets no PDF, which is this issue.
- The stale comment in `packages/finance-react/src/hooks/use-invoice-mutation.ts` documents the phantom and is wrong in both halves: `renderInvoice` does not emit `invoice.rendered` (only `bindInvoiceRendition` does), and no subscriber turns anything into a PDF.

mihaipxm:
Closes #4668.

## The gap

`renderInvoice` inserted a `pending` rendition and its own docblock said a background job would fulfil it. No such job existed — nothing in the repository read `invoice_renditions` where `status = 'pending'`. The row was an orphan: `?wait=true` polled it for its full 30s and returned 202, and the booking-confirmation notification held for a document that was never coming. `POST /invoices/{id}/generate-document` answered 501, because `createFinanceRuntime` never set `invoiceDocumentGenerator` and no deployment supplied one — `createPdfInvoiceDocumentGenerator` had been written, exported and tested, and its only callers were a unit test and the export checker.

So a deployment without an accounting app could not produce an invoice PDF at all.

## What decided the shape

The issue reads as "wire the existing generator into `createFinanceRuntime`". That is not possible: a graph runtime factory receives `getPort`/`hasPort` and no `getResource`. Resources reach only **selected provider factories** (`packages/runtime/src/index.ts`). Until finance declared a provider there was no seam through which a renderer could arrive, so the port is the mechanism, not ceremony.

## The four pieces

| | before | after |
|---|---|---|
| resources | none declared | `#resource.document-storage` + `#resource.document-renderer`, both required — an unfulfillable deployment now fails at graph resolution, not at a 501 |
| provider | an optional options field nobody set | `finance.invoice-document-provider`, conformance-tested, selected as `invoiceDocumentArtifact: "standard"` in `operator-standard` |
| trigger | none | an `invoice.issued` subscriber renders in-process |
| recovery | none — finance declared no jobs | `finance.invoice-document-renditions`, `every: "1m"`, `wakeup: true` |
| the miss | silent orphan | recorded on the row as `failed` with a reason |

The subscriber matters beyond latency: #4667 made the confirmation email wait for the invoice to resolve, so without an in-process trigger every confirmation would sit behind the job's cadence. Booking create cannot render inside its own transaction, so it writes the request and this turns it into a document.

**One deliberate divergence from the legal provider.** A repeated `put` with different bytes *replaces* rather than being rejected, because the operation key is one `invoice_renditions` row and re-rendering that row is exactly what a retry is. Rejecting the mismatch would strand the row behind any renderer whose output is not byte-identical run to run.

**The requested row is fulfilled in place** rather than superseded by a second `ready` row beside it — that orphan is what this replaced. `bindInvoiceRendition` keeps inserting; it serves the app-provided path, where bytes arrive without a request having been made.

**An invoice awaiting external allocation is skipped, not rendered.** When the series delegates numbering to an accounting app, the number is still `PENDING-…`; rendering locally would produce a fiscally invalid PDF naming a number no authority issued.

## Verified end to end, not just in tests

Booted `apps/operator` against Postgres in Docker and drove the real HTTP surface:

- `POST /invoices/{id}/render?wait=pdf` → **201**, `status: ready`, in 1.1s. Previously: a `pending` orphan and a 202 after the full wait.
- The artifact is a real **PDF 1.7, 96,164 bytes**, at the key the row points to, and its SHA-256 equals the `checksum` column exactly.
- `POST /invoices/{id}/generate-document` → **201**. Previously: 501.
- `GET /v1/admin/documents/files/…` → **200**, `application/pdf`, checksum matching the row.

That last one only works because of the second commit.

## Three further defects this turned up

Found while getting the finance integration suite honest; each is a case where the code discarded or refused something it should not.

1. **Private document downloads were 500 on every Node deployment.** The contract-document bundle pointed at `createContractDocumentApiModule` — the raw factory that takes its options as its first argument — instead of `createContractDocumentVoyantRuntime`, the wrapper that resolves them from the port. The graph handed its own context in as `options` and every request died destructuring `undefined`. That is the download target for contract PDFs and now invoice renditions.
2. **A reused idempotency key was a 500.** `ActionLedgerIdempotencyConflictError` reached the boundary unrecognized, so a client mistake the HTTP idempotency middleware already answers 409 for read as an infrastructure fault.
3. **A proforma could not share a number with its invoice.** Issuance keyed the action ledger on the invoice number alone, while `invoices_invoice_number_type_active_idx` is unique on (number, *type*) — the database deliberately permits the pair. The second document failed. Now keyed on (type, number).

Plus: conversion conflicts returned a sentence and threw away the pointer to the colliding invoice that the service had already resolved, so an operator could not open it. The 409 now carries `code`, `existingInvoiceId` and `existingInvoiceNumber`, and the route declares that shape.

## The finance integration suite

It had been failing on every run long enough to be read as "known red". `routes.test.ts` alone was **108 failing**; all six affected files are now green. Four mechanical causes and a tail of stale assertions, each checked against the change that moved the behaviour before being touched:

- #4100 removed `.default("draft")` from `bookings.status` — 106 of the 108 were fixtures omitting a now-required column.
- The harness mounted routes on a bare `Hono` with no error boundary, so every thrown validation/conflict became a default 500 and tests asserting 400/409 could never pass.
- `idempotency_keys` and the action-ledger tables were not truncated between tests, so a rerun matched a stored key with a different fingerprint and answered 409 to the *first* request.
- Assertions pinned shapes that moved: the series TypeID prefix is `invs`, and captured events are `EventEnvelope`s that gained `eventId`/`emittedAt`.
- CHECK-constraint tests matched the constraint name against `error.message`, which Drizzle's wrapper no longer carries — they passed on *any* failure, including a bad fixture. They now walk the cause chain.
- `POST /invoices` has no `invoiceType`; tests passing one were silently getting a plain invoice and asserting against it.
- #4050 made the public finance routes storefront-scoped, so all twelve 403s were the guard working, not the routes broken.
- A capability scoped to another booking is 403, not 401 — `ForbiddenApiError` deliberately separates "valid token, wrong subject" from "no token".

## Verification

- `pnpm verify:architecture` — passes, 99 chain links, including `verify:openapi-drift` and `verify:graph-conformance`.
- `biome check .` — 6,932 files, **0 errors** (23 warnings, 8 infos).
- `tsc -p tsconfig.build.json` and `tsconfig.typecheck.json` for finance — 0 errors; `turbo build` 34/34.
- `finance` — 79 files / 597 tests. `legal` — 26 files / 196 tests. `operator-standard` — 21 tests.
- Integration against Postgres: `routes` 147, `booking-create` 103, `currency-amount-checks` 17, `public-routes` 15, `invoice-document-fulfilment` 8 (new), `document-routes` 4, `unsynced-proforma-command` 4, `settlement-routes` 1 — **299 passing, 0 failing**.
- The new integration file is registered in the `db-integration` lane in `ci.yml`; that lane lists files by name, so an unregistered one would never run.
- Every pre-existing failure was confirmed against a pristine `origin/main` worktree with identical test files before being attributed or fixed.

## Not done

- The operator UI was not driven in a browser: the chrome-devtools MCP profile was held by another session for the whole run. The HTTP surface and the stored bytes were verified directly instead.
- No entry was added to `graph-conformance.json` — finance has none today, and the four declarations are covered by assertions in `voyant-manifest.test.ts`.

### DIFF
--- a/.changeset/invoice-document-provider.md
+++ b/.changeset/invoice-document-provider.md
@@ -1,0 +1,21 @@
+---
+"@voyant-travel/finance": minor
+"@voyant-travel/operator-standard": patch
+---
+
+Produce invoice documents on a deployment that has no accounting app.
+
+`renderInvoice` wrote a `pending` rendition that nothing in the repository ever
+fulfilled, and `POST /invoices/{id}/generate-document` answered HTTP 501 because
+no deployment supplied a generator. Finance now declares the `document-renderer`
+and `document-storage` resources it needs, offers a conformance-tested
+`finance.invoice-document-provider` port selected as `invoiceDocumentArtifact:
+"standard"` in `operator-standard`, renders in-process off `invoice.issued`, and
+carries a one-minute recovery job for everything the in-process path misses.
+
+A requested rendition is now fulfilled in place rather than orphaned beside a
+separate `ready` row. When no renderer is available the miss is recorded on the
+row as `failed`, so `?wait=true` and the booking-confirmation notification stop
+waiting on work that will never happen. An invoice whose number is still awaiting
+allocation by an installed accounting app is left alone rather than rendered
+against a placeholder number.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -387,6 +387,8 @@ jobs:
                 tests/integration/payment-disputes.test.ts
               pnpm --filter @voyant-travel/finance exec vitest run \
                 tests/integration/refund-settlements.test.ts
+              pnpm --filter @voyant-travel/finance exec vitest run \
+                tests/integration/invoice-document-fulfilment.test.ts
               pnpm --filter @voyant-travel/storefront exec vitest run \
                 tests/integration/payment-reconciliation-job.test.ts
               pnpm --filter @voyant-travel/bookings exec vitest run \
--- a/apps/operator/tests/agent-journey-real-surface.test.ts
+++ b/apps/operator/tests/agent-journey-real-surface.test.ts
@@ -258,6 +258,19 @@ function formatReport(scores: readonly DiscoveryScore[]): string {
  * winning back, cap the default `search_tools` limit rather than narrowing the
  * vocabulary — the vocabulary is what makes the surface findable at all.
  *
+ * RAISED 48,000 → 49,000: measured 48,436, and **not by the change that
+ * surfaced it**. `origin/main` measures the identical 48,436 — the number is
+ * byte-for-byte the same on a pristine checkout — so this is drift that
+ * accumulated since the 43,730 recorded below. Main reads green because the
+ * push build gets a Turbo cache hit on an unchanged `operator` test task, while
+ * a PR runs `pnpm test --affected` and actually executes it: any PR that makes
+ * operator affected is the first to run the assertion, and pays for drift it
+ * did not cause. The report still shows the narrowed column at 32,613 (-33%),
+ * so the per-call lever is intact and nothing ballooned; the broad
+ * `bookings_query` line (9,229 x2) is the largest and is what the
+ * `search_tools` limit above would cut. Raised to the measurement plus ~1%
+ * rather than to a round number, so it keeps tripping.
+ *
  * The layered read projection (voyant#3932) then collapsed the ~133 reads into
  * ~24 `<domain>_query` tools: a journey now describes the query tool for the
  * record it wants, which advertises the union of its resources' compact INPUT
@@ -267,7 +280,7 @@ function formatReport(scores: readonly DiscoveryScore[]): string {
  * drift (a broad `search_tools` over the many booking WRITE tools is the largest
  * remaining line), while still tripping if a describe payload balloons again.
  */
-const DISCOVERY_TOKEN_CEILING = 48_000
+const DISCOVERY_TOKEN_CEILING = 49_000
 
 /**
  * This hook composes the whole selected graph and runs every journey, which
--- a/packages/accommodations/tsconfig.build.json
+++ b/packages/accommodations/tsconfig.build.json
@@ -1049,7 +1049,11 @@
       "@voyant-travel/finance/checkout": ["../finance/dist/checkout-service.d.ts"],
       "@voyant-travel/finance/checkout-routes": ["../finance/dist/checkout-routes.d.ts"],
       "@voyant-travel/finance/checkout-validation": ["../finance/dist/checkout-validation.d.ts"],
+      "@voyant-travel/finance/invoice-document-job": ["../finance/dist/invoice-document-job.d.ts"],
       "@voyant-travel/finance/invoice-fx": ["../finance/dist/invoice-fx.d.ts"],
+      "@voyant-travel/finance/invoice-issued-subscriber": [
+        "../finance/dist/invoice-issued-subscriber.d.ts"
+      ],
       "@voyant-travel/finance/linkables": ["../finance/dist/linkables.d.ts"],
       "@voyant-travel/finance/order-payment-sessions": [
         "../finance/dist/order-payment-sessions.d.ts"
--- a/packages/accommodations/tsconfig.typecheck.json
+++ b/packages/accommodations/tsconfig.typecheck.json
@@ -1050,7 +1050,11 @@
       "@voyant-travel/finance/checkout": ["../finance/dist/checkout-service.d.ts"],
       "@voyant-travel/finance/checkout-routes": ["../finance/dist/checkout-routes.d.ts"],
       "@voyant-travel/finance/checkout-validation": ["../finance/dist/checkout-validation.d.ts"],
+      "@voyant-travel/finance/invoice-document-job": ["../finance/dist/invoice-document-job.d.ts"],
       "@voyant-travel/finance/invoice-fx": ["../finance/dist/invoice-fx.d.ts"],
+      "@voyant-travel/finance/invoice-issued-subscriber": [
+        "../finance/dist/invoice-issued-subscriber.d.ts"
+      ],
       "@voyant-travel/finance/linkables": ["../finance/dist/linkables.d.ts"],
       "@voyant-travel/finance/order-payment-sessions": [
         "../finance/dist/order-payment-sessions.d.ts"
--- a/packages/catalog/tsconfig.build.json
+++ b/packages/catalog/tsconfig.build.json
@@ -998,7 +998,11 @@
       "@voyant-travel/finance/checkout": ["../finance/dist/checkout-service.d.ts"],
       "@voyant-travel/finance/checkout-routes": ["../finance/dist/checkout-routes.d.ts"],
       "@voyant-travel/finance/checkout-validation": ["../finance/dist/checkout-validation.d.ts"],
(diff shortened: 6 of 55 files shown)


mihaipxm:
All three review findings were real and are fixed in `02269da51c`, plus the CI failure.

**P1 — custom fields on background rendering.** Correct, and worse than described: the resolver was defined inline in `createFinanceRuntime`, so there was nothing for the other paths to reuse even if they had tried. Extracted as `createInvoiceCustomFieldsResolver`, and both graph runtimes now resolve `custom-fields.runtime` and pass it. A template using `{{customFields.*}}` no longer renders differently depending on which path produced the document.

**P1 — hold the claim across the storage write.** Correct. `pg_advisory_xact_lock` ends with its transaction, well before `render`/`put`. The claim is now written *into the row* while the lock is held, so it outlives the transaction: a second caller sees a live claim and never renders at all. The closing CAS additionally requires the claim to still be ours, so a claim that expired mid-render cannot clobber whoever took over. The lease is five minutes — a process that dies mid-render must not strand the row, which would recreate the orphan this PR removed — and a caller that decides *not* to render hands it back explicitly rather than sitting on it.

Covered by a test that races two `fulfilInvoiceRendition` calls at one row and asserts one render, one stored object, and `fileSize` equal to the bytes actually in the store.

**P2 — preserve non-blocking behaviour.** Correct, and thank you for the doc reference — gated on `waitRequest.mode !== "none"`.

Worth noting for reviewers: this fix alone would have regressed the operator's Generate/Regenerate button, which sends no `wait`. So `use-invoice-mutation` now opts in explicitly (`wait: "pdf"`, overridable), which is what the contract is for — a person who just pressed a button is the caller a bounded wait exists for. Its docblock also claimed the route emitted `invoice.rendered` and that a subscriber turned it into a PDF; neither was ever true, and that gap is what #4668 is.

---

**CI** was one root cause, not three: `tests` failed → `checks` failed → `db-lanes` skipped → `db-checks` failed on `needs.db-lanes.result == "success"`.

The `tests` failure was `agent-journey-real-surface`'s discovery-token ceiling, and it is **not this branch's growth**: a pristine `origin/main` worktree measures the identical 48,436. Main reads green because its push build gets a Turbo cache hit on an unchanged `operator` test task, while a PR runs `pnpm test --affected` and actually executes the assertion — so any PR that makes operator affected is the first to run it and pays for drift it did not cause. Raised 48,000 → 49,000 (measurement + ~1%, so it still trips) with that recorded next to the constant, per its own "raise it only with a recorded reason after reading the printed report". The narrowed column is still 32,613 (-33%), so nothing ballooned; the broad `bookings_query` line is the largest and the `search_tools` limit named in that comment is the lever if someone wants it back.

Verified locally: 8 finance integration files green (303 tests, including the three new ones), finance 597, finance-react 70, legal 196, operator-standard 21, `verify:architecture` 99 links exit 0, `biome check .` 0 errors, builds 0 `error TS`.

PR #4710, opened 2026-08-15
Speak the operator's language on the catalog browse surfaces

mihaipxm:
The catalog browse surfaces were showing the search index's vocabulary rather than the catalogue's. Each item below is one of those.

### Every schedule entry was a "departure"

`scheduleTerm` (`session` | `occurrence` | `departure`) is resolved once in `@voyant-travel/inventory` and **already carried on the catalog document** (`service-catalog-plane.ts`), so every surface can agree on the noun. The card ignored it and hardcoded `{count} departures` — so a timed Activity and a one-night gig both had "departures".

Now a sixty-minute sail has **sessions**, an Event has **dates**, a Tour keeps **departures**. Unknown or absent term falls back to departures.

### A timed product hid the thing that distinguishes it

An instant carries an hour; the card printed only the calendar day, so a 09:00 and an 18:00 sailing looked identical. The time is now shown, and hovering the footer gives the same instant in **both** frames:

```
There: 25 Sept, 12:30 (Pacific/Auckland)
Yours: 25 Sept, 01:30 (Europe/Bucharest)
```

It offers nothing when there is nothing to convert — a bare `YYYY-MM-DD` has no clock reading, and a reader already in the departure's zone would just see the same line twice. The trigger is a real `<button>`, so the second frame is reachable by keyboard and not mouse-only.

### `1d / 0n`

Zero nights is the absence of an overnight, not a measurement. A single-day product reads `1d`; `3d / 2n` is unchanged.

### `Free_sale`

The rail leaned on CSS `capitalize`, which only touches the first letter of each *word* and does not treat `_` as a word break. Codes are now read as sentences — "Free sale", not "Free_sale" and not Title Case. Values that already carry their own casing (`TUI`) are left alone, and fields with a real vocabulary keep their `formatValue`.

### The page offered back the facet it pins

The Tours page **is** `familyCode = tour`. `CatalogBrowsePage` merged that lock into the reader's filters, so the rail rendered it as a checked box with a Clear link that could not clear anything — the lock re-applied on the next render — while spending rail space restating the page title. `hiddenFilterFields` on `CatalogPage` drops it. The filter still applies; it is just no longer presented as the reader's to change.

### Booking mode was a filter

It is the integration mechanic a product is sold through, derived rather than authored (ADR-0010). Nobody browsing a catalogue is looking for "the date_time ones".

### Family pages had no search at all

They embed the browse grid with `embedded`, which sets `hideSearchInput` so there is one search per page — but nothing supplied that one. An operator with a few hundred tours could only page through them.

`CatalogSearchInput` now serves both surfaces. Extracting it also **removed** the duplicated typing-buffer/debounce/last-emitted bookkeeping from `CatalogSearchPage` (three effects and two refs down to one callback), so there is one place where "don't push a history entry per keystroke, don't clobber the caret" is handled.

### Taglines restated the heading

"Products in the Activity family." under a heading reading **Activities**. They now say what belongs in the family, drawn from the seeded family descriptions. en + ro both updated.

### Verification

- `catalog-react` **105 tests pass, up from 85** — 20 new, covering each item above
- `tsc -p tsconfig.typecheck.json` (includes tests) → exit 0, 0 `error TS`
- `biome check .` from repo root → 0 errors, back to main's 23 warnings / 8 infos
- rebased on `main` after #4708 and re-verified

**Not browser-verified.** Every change here is unit-tested at the projection/config level, but I have not rendered these pages against seeded data. Worth a look at real density before merge, particularly the tooltip placement on the card and the search bar's position in the family header.


mihaipxm:
Both review comments were correct and are fixed in `93a8291601`, along with the `architecture` CI failure.

**`nextDepartureAt` vs `nextDepartureDate`** — confirmed against `catalog-policy-departures.ts`, which states the convention outright: `…At` is an ISO instant, `…Date` is a bare local calendar date. Both the hour and the tooltip were reading the bare field, so both were dead in production while their tests passed — the tests hand-constructed an instant-valued `nextDepartureDate` that no projection emits. Exactly the failure mode AGENTS.md describes: a green from a path production does not take.

Beyond swapping the field, the frame is now decided by **which field the value was read out of** rather than by the shape of the string. Sniffing would mean a malformed document silently picks up a time of day in a zone nobody declared. A new test pins that: an instant-shaped value sitting in `nextDepartureDate` renders no time.

This also surfaced a third thing neither of us had flagged — cruises project `earliestDepartureCached` as a `date` column, so there is no instant on that vertical at all. That card now declares no tooltip rather than one that could never fire.

**Hidden locks in selection accounting** — confirmed: `selections.facets` carries the merged lock, so `hasFacetSelections` was permanently true and every fresh family page offered "Clear all". `hiddenFilterFields` now reaches `CatalogTabPanel`, which excludes those fields from the count. Clearing preserves them rather than dropping them, since dropping wrote a filter-less URL that the surface re-merged on the next render — the grid would flash the unscoped result set on the way back.

**CI** — `architecture` failed on `Intl.DateTimeFormat()` with no locale in `resolveViewerTimeZone`. `resolvedOptions().timeZone` is a property of the environment rather than the locale, so binding it explicitly changes nothing except that the rule keeps having no exceptions. `verify:architecture` now passes locally end to end; it is not in the husky hook, which is why it reached CI.

113 tests (up from 105), typecheck-including-tests exit 0, root biome 0 errors.

PR #4711, opened 2026-08-15
Ask staff whether to notify the customer, not whether to suppress

mihaipxm:
The status-change dialog carried a switch labelled **"Don't notify the customer"**, off by default — a box you tick to make less happen. The booking journey's equivalent already asks the question positively (**"Notify traveler"**, on by default) and maps it to the wire flag itself, so the same decision had two opposite presentations depending on where an operator made it.

| | before | after |
|---|---|---|
| status dialog label | "Don't notify the customer" | "Notify the customer" |
| default | off | on |
| wire | `suppressNotifications: values.suppressNotifications \|\| undefined` | `suppressNotifications: values.notifyCustomer ? undefined : true` |

## The domain flag deliberately stays negative

Only the dialog flips. `suppressNotifications` and `bookings.notifications_suppressed` are unchanged, because the negative default is the fail-safe one:

- A call site that forgets the flag sends a redundant email — noisy, visible, recoverable. A forgotten opt-in leaves the customer never told, and silence fails silently.
- Every downstream consumer in `notifications` is a skip-guard (`service-reminders.ts`, `service-reminder-events.ts`, `service-reminder-stage-runs.ts`, `subscriber-runtime.ts`). Inverted, each would have to read a missing or unhydrated field as "stay silent".
- `updateBookingSchema` types the column as `z.literal(true)` — a deliberate monotonic latch. Inverted it reads "`notificationsEnabled` may only ever be set to `false`", which is the same latch spelled backwards and looks like a bug.

## Both toggles now disclose the latch

Switching notifications off doesn't just skip this action's email. It latches `bookings.notifications_suppressed`, which nothing can clear — the booking is silent for good, future reminders included. This holds on all three paths: cancel (`service-core.ts:3267`), override-status (`:3661`), and create (`:2716`).

The old copy described it as confirming "silently", which reads as a one-time choice about the action in front of you. Both surfaces now say it is permanent, in en and ro.

## Verification

- `turbo run typecheck --filter=@voyant-travel/bookings-react` — 68/68 tasks, exit 0, zero `error TS`
- `vitest run tests/unit` — 30 files / 208 tests pass, including 3 new
- `biome check .` from the repo root — 0 errors; no diagnostics on changed files
- `pnpm i18n:check` — passes, en+ro parity kept

The new test pins the polarity rather than just asserting current behaviour: flipping the default back to off makes `defaults to notifying the customer` fail, and restoring it makes it pass. Verified both directions rather than trusting the green.

## Not addressed here

`status-dispatch.ts:41` honours `suppressNotifications` only for `cancelled`/`confirmed`, and `suppressLifecycleEvents` only for `confirmed`. The dialog hides the toggle elsewhere so the UI agrees today, but the admin API and MCP tool accept the flag on other transitions and silently ignore it. Separate change.

### DIFF
--- a/.changeset/notify-customer-toggle-polarity.md
+++ b/.changeset/notify-customer-toggle-polarity.md
@@ -1,0 +1,11 @@
+---
+"@voyant-travel/bookings-react": patch
+---
+
+Ask staff whether to notify the customer, rather than whether to suppress the notification.
+
+The status-change dialog carried a switch labelled "Don't notify the customer", off by default — a box you tick to make less happen. The booking journey's equivalent already asks the question positively ("Notify traveler", on by default) and maps it to the wire flag itself, so the same decision had two opposite presentations depending on where an operator made it. The status dialog now matches: `notifyCustomer` is on by default and sends `suppressNotifications: true` only when switched off.
+
+The domain flag stays negative. `suppressNotifications` is the safe default at the boundary — a call site that forgets it sends a redundant email, where an opt-in flag that is forgotten leaves the customer never told, and every downstream consumer in `notifications` is a skip-guard that would have to read a missing field as "stay silent".
+
+Both toggles now also disclose what switching notifications off actually does. Turning it off latches `bookings.notifications_suppressed`, which `updateBookingSchema` types as `z.literal(true)` — nothing can clear it, so the booking is silent for good, future reminders included. The old helper text described it as confirming "silently", which reads as a one-time choice about this action.
--- a/packages/bookings-react/src/components/booking-detail-page.tsx
+++ b/packages/bookings-react/src/components/booking-detail-page.tsx
@@ -639,6 +639,7 @@ export function BookingDetailPage({
         onOpenChange={setStatusDialogOpen}
         bookingId={id}
         currentStatus={booking.status}
+        notificationsSuppressed={booking.notificationsSuppressed === true}
       />
 
       <BookingCancellationDialog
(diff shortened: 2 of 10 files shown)


mihaipxm:
Thanks — one real catch, two I'm not acting on. Detail below.

### P1 "Do not show notify enabled for latched-silent bookings" — correct, fixed

Confirmed against the tree: `packages/bookings/src/routes-admin.ts:1293` returns `notificationsSuppressed`, but `bookingRecordSchema` in `packages/bookings-react/src/schemas.ts` never parsed it, so the dialog had no way to know. A booking latched silent by an earlier action showed a checked "Notify the customer" and submitted no suppression flag, promising a send that the service would discard.

- `notificationsSuppressed` added to `bookingRecordSchema` (optional — responses predating the field omit it)
- `StatusChangeDialog` takes it as a prop; when latched the switch is off and `disabled`, with helper copy saying the booking was silenced earlier and it cannot be undone (en + ro)
- `booking-detail-page.tsx` passes it through
- New test pins it, and I verified it is not vacuous: reverting the fix makes `shows the toggle off and disabled when the booking is already silenced` fail

Worth noting the same false promise existed before this PR — "Don't notify" sat unchecked on a latched booking too. The positive framing made it louder, which is how it surfaced.

### P1 "Attribute the commit to the human author" — incorrect

`git show --format=fuller 57984928df` reports:

```
Author:     Mihai Ungureanu <mihai@pixelmakers.com>
Commit:     Mihai Ungureanu <mihai@pixelmakers.com>
```

Both trailers are the human author, and there is no `Co-authored-by` or generated-with footer. No change needed.

### P2 "Preserve the published override keys in a patch release" — does not apply

`packages/bookings-react/package.json` sets `"private": true` (the `publishConfig.access` above it is inert), and the package is absent from `scripts/checks/manifests/public-surface.json`. It is not on npm, so there are no external consumers holding `BookingsUiMessageOverrides` against the old key path. In-repo, `statusChangeDialog.fields.suppressNotifications` and its helper had exactly one reader — the dialog itself — and no override supplies them. Patch is right.

PR #4718, opened 2026-08-16
Restore the content catalog detail pages were dropping

mihaipxm:
Fixes the report that individual catalog detail pages look broken or are missing information — e.g. `/catalog/activities/prod_01kz996xesegnv4vd6aescjnhf` on sandbox.

### The page was rendering a record it never fetched

`CatalogVerticalDetailPage` built its hit with an empty field map:

```ts
const hit = { id, score: 0, document: { id, fields: {} } }
```

Entered by id there is no result row to carry the index projection, so price, offers, status, categories and destinations were all absent and the **entire Attributes tab disappeared** (`hasAttributes` is computed from those fields). The same record showed far more when opened as a sheet from the list than on its own page. It now reads the indexed document by id.

The supplier formatter is also held by ref, so the supplier directory settling no longer rebuilds the enrichment fetchers and re-requests the record — one page load was issuing the content route three times.

### Sourced cruises were degrading to blank in four separate ways

- **Adapter resolved by connection alone.** A channel registers several adapters per connection and keys them apart by suffixing the registry key (`<connectionId>:cruises`), so resolving by connection returned the connection's *generic* adapter, which carries no `cruiseAdapter`. Every connection-scoped cruise reported no cabin pricing at all. Now resolved by its own `source_kind` scoped to the connection, falling back to the bare connection id.
- **The synthesizer read the wrong casing.** It read the snake_case names of the content shape it *produces* rather than the projection's camelCase keys. Those overlapped on `id`/`name`/`status` and nothing else, so the fallback rendered blank even when discovery had captured the data.
- **`source_provider` held a connection id.** The shim read `sourceRef.provider` while Voyant Connect writes `providerKey`, so a raw `conn_…` string surfaced as the cruise line. Now stamps the provider key, and projects cruise line, ship and port display names alongside their faceted ids.
- **A failing adapter 500'd the route.** `resolveCruiseRow` in `@voyant-travel/connect-adapter` throws `Connect cruise content not found` for cruises discovery has already indexed, so every concurrent cruise detail open on sandbox became a 500. We hold a durable sourced-entry projection and §3.6 defines the synthesizer as exactly that degraded read, so it now degrades instead of throwing. The failure is reported through a new `onContentFetchError` option (default `console.warn`) so an upstream outage stays visible rather than silently degrading every cruise to a stub.

### Owned products

Fall back to an itinerary-day image for the hero when there is no product-level image, instead of reporting no hero while the same image sits in `content.media`.

### Verification

- `catalog-react` 89, `inventory` 601 (+71 skipped), `cruises` 284 (+7 skipped) — all pass
- `tsc -p tsconfig.typecheck.json` on `catalog-react` → exit 0, 0 `error TS`
- `pnpm verify:architecture` → exit 0
- `biome check .` from repo root → 0 errors

### Provenance

This work was written in an earlier session, never pushed, and left on a local branch based on stale main. It has been rebased onto current main and re-verified end to end rather than rewritten.

### DIFF
--- a/.changeset/catalog-detail-missing-info.md
+++ b/.changeset/catalog-detail-missing-info.md
@@ -1,0 +1,49 @@
+---
+"@voyant-travel/catalog-react": patch
+"@voyant-travel/cruises": patch
+"@voyant-travel/inventory": patch
+---
+
+Restore the content catalog detail pages were dropping.
+
+Resolve a sourced cruise's adapter by its own `source_kind` scoped to the
+connection before falling back to the bare connection id. A channel registers
+several adapters per connection and keys them apart by suffixing the registry
+key (`<connectionId>:cruises`), so resolving by connection alone returned the
+connection's *generic* adapter. That adapter carries no `cruiseAdapter`, so
+`getCruiseSailingPricing` reached through it for nothing and every
+connection-scoped cruise reported no cabin pricing at all.
+
+Read the projection's own camelCase keys in the cruise content synthesizer. It
+read the snake_case names of the content shape it produces, which overlapped the
+shim's projection on `id`/`name`/`status` and nothing else, so the fallback
+rendered blank even when discovery had captured the data.
+
+Stamp the provider key rather than the connection id as a sourced cruise's
+`source_provider`, and project the cruise line, ship and port display names
+alongside their faceted ids. The shim read `sourceRef.provider` while Voyant
+Connect writes `providerKey`, so the fallback fired every time and a raw
+`conn_…` string surfaced as the cruise line on the detail page.
+
+Read the indexed document by id on the URL-addressable vertical detail pages.
+Entered by id there is no result row to carry the index projection, so price,
+offers, status, categories, destinations and the whole Attributes tab were
+dropped — the same record showed far more when opened as a sheet from the list.
+The supplier formatter is now held by ref so the supplier directory settling no
+longer rebuilds the fetchers and re-requests the record (one page load issued the
+content route three times).
+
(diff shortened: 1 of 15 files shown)


mihaipxm:
Valid, and fixed in `8b3dfe0a71`.

Traced the whole chain rather than taking it on trust:

1. `catalog-enrichment-mappers.ts:236` calls `formatSupplier(...)` **once**, when the content is mapped, and freezes the result into `enrichment.supplier`.
2. `catalog-vertical-detail-page.tsx` rendered that snapshot directly.
3. The ref-stable callback has `[]` deps, so when `useSuppliers` settles there is no identity change, no fetchers rebuild, and no refetch — nothing remaps the frozen string.

So on any load where the content request beats the supplier directory, the header kept showing the raw `sup_…` forever. This PR had traded a triple-fetch for that, which is a bad trade in one direction and a good one in the other.

**Fix: resolve at render instead of only at map time.** `formatSupplier` is a map lookup that returns its input on a miss, so applying it to an already-resolved name is a no-op — which makes re-resolving safe and idempotent, and keeps the fetchers independent of the directory so the triple-fetch stays fixed.

Chose this over remapping the enrichment on directory change: the enrichment is a snapshot of a fetched document, and mutating it in place to correct one derived label makes it something else. The label is presentation, so it is resolved where it is presented.

Four tests added covering the load-order race directly — directory-first, content-first, idempotence over an already-resolved name, and a null supplier staying null. `catalog-react` 93 tests pass, typecheck exit 0, biome 0 errors.

PR #4722, opened 2026-08-16
Give ships a surface of their own

mihaipxm:
**Stacked on #4718** — review that first; this branch contains its commit.

You asked for ships to exist separately from cruises. They already did, everywhere except the screen.

`cruise_ships` carries IMO, deck and cabin counts, guest and crew capacity, length, cruising speed, year built and refurbished, gallery and deck plan. Behind it sit admin and public REST routes (`routes-ships.ts`), editorial overlays (`writeCruiseShipOverlay`), external-ship ingestion, a field policy and a document builder. What was missing was any way for an operator to look at one — no nav entry, no route, no page.

### The surface

`/catalog/ships` and `/catalog/ships/$shipId`, contributed by a new `@voyant-travel/cruises-react/admin` extension and declared on the cruises module's admin facet.

A vessel is referenced by many sailings and outlives all of them, so it reads on its own page rather than as a tab inside whichever cruise happened to be open. That is also the page that says **"About this ship"** — which a cruise detail page has no business claiming about a hotel.

### Why a plain list and not a catalog vertical

There is no ships collection in the search index to facet against. Standing one up to power a reference list of a few hundred rows would be a lot of machinery for no gain, so this reads the existing paginated `GET /ships` route. Search is server-side on name and IMO.

### Sparse rows

A sourced vessel often arrives with a name, a type and nothing else. An absent specification is omitted entirely rather than drawn as a dash — otherwise the page fills with placeholders and buries the two facts that did come through. A `0` guest capacity is treated as an unfilled column, not a capacity, and the `numeric` columns' trailing zeros are dropped (`294.00` → `294 m`) without touching real digits.

### Two things the architecture chain caught, both worth keeping

- **`admin.runtime` was missing.** Exposing `@voyant-travel/cruises-react/admin` without declaring it on the module made `verify:graph-conformance` fail, then the operator graph build rejected the routes for having no runtime reference. Both are now declared, and `apps/operator` builds its artifacts with the Ships routes in place.
- **The destinations test was not typechecked by any CI job.** It is a *compile-time* proof that the `AdminDestinations` augmentation and a resolver map agree — and a proof nothing typechecks proves nothing. `cruises-react` now typechecks its tests via the #4243 tsconfig split (`tsconfig.build.json` stands alone, so `dist` is unaffected), which shrinks `verify:typecheck-coverage`'s baseline from **56 packages to 55**.

The host resolver for `cruiseShip.detail` / `cruiseShip.list` lives outside this tree, same as `catalog.detail` — until it is registered, `useAdminHref` yields `"#"` and `openHrefInNewTab` declines to open it, so the list degrades to inert rather than navigating somewhere wrong.

### Verification

- `cruises-react` 18 tests (12 new), `cruises` 284, `i18n` 13 — all pass
- `tsc -p tsconfig.typecheck.json` → exit 0, 0 `error TS`, now covering `tests/**`
- `pnpm verify:architecture` → **exit 0** (full chain)
- `apps/operator` `prepare:verify` → exit 0, artifacts generated
- `biome check .` → 0 errors
- en + ro copy for every new string

### Not in this PR

Ship **authoring** — the overlay and translation machinery exists and this page is read-only. Worth its own PR so the read surface can land and be looked at first.


mihaipxm:
All three confirmed against the code and fixed in `51990e675f`.

### P1 — the page was in its error state 100% of the time

Correct, and worse than a page-size nit. `shipListQuerySchema` (`packages/cruises/src/validation-cabins.ts:51`) caps `limit` at **100**; the surface asked for 200. Validation **rejects rather than clamps**, so every request 400'd and the browse page never once rendered a ship.

The comment above the constant asserted the route capped at 200 — I had read the shared `paginationSchema` from `@voyant-travel/types` rather than the cruises-specific schema, and wrote the wrong fact down confidently next to the value it justified.

Now 100, with the comment saying *why* the number matters (it rejects, it does not degrade). Added `tests/unit/ships-search.test.ts` pinning 101 and 200 as rejected and 100 as accepted — the assertion that would have caught this without a browser.

Also: a fleet longer than one page now reads **"Showing 100 of N ships — search to narrow it down."** Printing the bare total over a truncated grid claims to be the whole fleet.

### P2 — the Romanian copy would have shipped and never rendered

Correct. Without `withAdminRouteMessagesProvider`, `useCruisesUiMessagesOrDefault()` falls back to its hard-coded English, so a Romanian operator would get a localized nav label and an English page behind it. Wrapped the selected extension the way `catalog-react` does.

### P3 — IMO search was advertised, not implemented

Correct: `listShips` matched `ilike` on `name` only. Rather than withdraw the offer in the placeholder, I made the predicate true — `or(ilike(name), ilike(imo))`. An IMO is how a vessel is identified unambiguously across cruise lines, so it is the more precise of the two things the box offers and the one worth keeping.

### Verification

- `cruises` 287 passed (+3 new), `cruises-react` 18 passed
- `tsc -p tsconfig.typecheck.json` → exit 0, 0 `error TS`
- `pnpm verify:architecture` → exit 0, full chain
- `biome check .` → 0 errors

Worth stating plainly: all three of these are things a browser would have surfaced immediately, P1 within seconds. Every check I ran was green while the feature did not work at all.

mihaipxm:
This merged into `fix/catalog-detail-missing-info` rather than `main`, and that base had already gone to main via #4718 nineteen minutes earlier — so the squash `a2298c23bb` sits on a branch nothing merges from and the Ships surface never reached `main`. Replayed onto main in #4759.

PR #4725, opened 2026-08-16
Show where a supplier is published, on the supplier's own page

mihaipxm:
Targets `main` directly — independent of #4718 and #4722.

## Nothing new was built. The control moved.

Supplier-level channel publication already exists end to end, and I want to be explicit about that because I initially reported it as missing:

- `effectivePublicationReasonSchema` carries **`supplier_decision`**, with `product_decision` overriding it and `default_deny` underneath — precedence covered by `publication-resolver.test.ts`
- full CRUD on `/supplier-publications`, plus `/preview` and `/publications/effective`
- `channel_publication_reindex_intents` with `product | supplier | source | catalog` scopes
- enforcement at **index time**: `isOwnedProductStorefrontListable` returns false without a channel, and `IndexerSlice` is `{vertical, locale, audience, market, channel?}`

So this is an indexing decision, not a read filter — publication decides which channel slices a record is indexed into at all.

It was reachable from exactly one place: **Channels → row menu → Publication**, a per-channel sheet. That is the wrong way round for the question an operator arrives with. "Stop putting this supplier on the website" is formed while looking at a supplier. Same authority, same endpoints; this only puts the control where the intent forms.

## What the page says

Every channel is listed, **including ones with no rule**, because "where does this supplier show up?" cannot be answered by a list of only the decided channels.

State is three-valued, and that is the substantive design call: **undecided is not excluded**. The resolver's default is deny, but a *product* rule can still publish an individual product from a supplier nobody has ruled on. Collapsing the two would report a supplier as blocked when nothing about it was ever decided.

A non-publishing channel reports as inactive, which is what the resolver answers before it looks at a rule at all.

## A bug this PR fixes in its own first draft

The model originally read `channel.active === false`. `channelRecordSchema` has **no `active` field** — it carries `status: active | inactive | pending | archived`. The property was `undefined` for every real channel, so the inactive state could never fire against live data, and the unit test passed anyway **because it hand-built the input**. That is the same "green from a path production does not take" that AGENTS.md warns about, and it is only caught here because I went and read the schema rather than trusting the test.

Now modelled on the real enum, with a case per non-publishing status.

## Verification

- `distribution-react` **55 tests pass** (13 model tests, up from 10)
- `pnpm verify:architecture` → exit 0, full chain
- `biome check .` → 0 errors
- **Typecheck could not complete locally.** `tsc` on this package exceeded 10 minutes repeatedly under parallel-worktree load (load avg ~10), on both the typecheck and build projects. Rather than claim it, I verified the two shapes the change depends on by reading the schemas directly — `channelRecordSchema.status` (which is how the bug above surfaced) and `channelSupplierPublicationRecordSchema` (`id`/`channelId`/`supplierId`/`decision`, all confirmed). CI's typecheck lane is the gate for the rest.
- Two unrelated component tests failed once under load and passed on a clean re-run; not from this change.
- en + ro copy for every new string.

## Not in this PR

The product-side half — "hidden on Website — supplier decision" on a product, from `/publications/effective`. That needs `inventory-react` to depend on `distribution-react`, which it currently does not, so it brings a new cross-package dependency plus dep-path and dist-config regeneration. Worth its own PR.

**Not browser-verified.**


mihaipxm:
Both valid and fixed — plus the CI failure, which was the typecheck I could not complete locally.

### CI: `build` / `build-graph`

`error TS2339: Property 'detail' does not exist on type 'SuppliersUiMessages'`. The supplier page's copy lives under `supplierDetailPage`, which is where I added the new block — but the component read `messages.detail.publication`. Fixed at both call sites, and I re-checked that the block is nested under `supplierDetailPage` in `messages.ts`, `en.ts` and `ro.ts`.

Worth stating: `tsc` on this package exceeds ten minutes on my machine and never completed across several attempts, so CI's `build-graph` (~70s) is what caught it. I flagged that gap in the PR body rather than claiming a pass, and this is exactly the miss it predicted.

### P2 — preview and confirmation before mutating

Correct, and the more serious of the two. A supplier rule is a bulk action over every product that supplier provides, and the channel sheet already treats it as one: `previewSupplier` → show `affectedProductCount` → explicit confirmation → `upsertSupplier`. This section wrote on the first click, so one press could reindex and flip visibility for an arbitrarily large catalogue with nothing shown first.

That is a safeguard regression relative to the flow this PR is supposed to be relocating, not adding to — the whole claim was "same authority, same endpoints". Now the same preview-and-confirm, inline on the channel row: pressing Publish/Hide previews, the row shows **"Affects N products."**, and the write only happens on Confirm.

### P2 — rules page smaller than the channel page

Correct. `useSupplierPublications` took the endpoint's default of 50 while the channel query asked for 100. A rule falling off that first page joins as "no rule" and renders as **"Not decided"** — a decided channel reported as undecided, directly next to a button that would overwrite it.

Rules are now fetched at 200 (the endpoint's cap) against a 100-channel ceiling, so the join cannot miss rather than being unlikely to. A channel list longer than one page now says so, instead of implying this supplier is absent from the channels it did not show.

### Verification

- `distribution-react` 55 tests pass
- `biome check .` → 0 errors
- `verify:architecture` → exit 0 (run before these edits; the change since is component-level)
- Typecheck: still cannot complete locally — CI is the gate. en + ro copy added for the new strings.

PR #4738, opened 2026-08-16
Refuse an over-long billing value at the Session write path, and stop a settled payment being stranded by the Session clock

### LINKED ISSUE #4733: A commit that fails after capture wedges the Booking Session permanently: quoting is refused as payment_in_flight, committing needs a fingerprint only a quote can give
## Summary

A commit that fails after the payment has been captured leaves the Booking Session permanently unrecoverable. There is no sequence of public API calls that can settle it afterwards, because the two routes needed to retry are mutually exclusive:

- `POST /v1/public/catalog/booking-sessions/{id}/quote` → `{"kind":"rejected","error":{"kind":"payment_in_flight","nextAction":"await_payment_outcome"}}`
- `POST /v1/public/catalog/booking-sessions/{id}/commit` → `400`, `requirementsFingerprint` is required

`requirementsFingerprint` is only ever produced by a quote response. So while a payment is outstanding, quoting is refused and committing is impossible. The money is captured, no booking exists, and the session cannot be advanced by anything the engine or the shopper can call.

## Reproduction

1. Reach the payment step with a value that the commit schema rejects but the session `PATCH` accepts (see the companion issue on validation asymmetry — a 25-character `address.postal` does it, `contactPostalCode` is capped at 20).
2. Pay by card. `payment_sessions.status` becomes `paid`.
3. The settlement commit answers `400 invalid_request`.
4. Every subsequent attempt: `quote` → `payment_in_flight`, `commit` → missing fingerprint.

Observed on a managed tenant, runtime `0.82.3`.

## Second trap: the correction is rejected as a replay

Correcting the offending field is the obvious escape, and it does not work. The update idempotency key is derived from the session revision:

```
idempotency-key: engine-update:<sessionId>:<revision>
```

The revision only advances on a successful mutation, and the failing commit never advances it. So the corrective `PATCH` reuses the key of the first (bad) `PATCH` at that revision and is answered:

```json
{"kind":"rejected","error":{"kind":"idempotency_conflict"}}
```

with `200 OK`. The original payload is replayed and the bad value stays. **The only escape from the failure requires an edit that the failure itself makes impossible.** Confirmed by watching the client send `"postal":"013912"` and the stored state remain `"Strada Aviator Popisteanu"` across a dozen attempts; a direct write to `booking_sessions.state_payload` unblocked it and the very next commit returned `200`.

## Third trap: the settlement window is minutes

In one live case the payment session expired **3 minutes 57 seconds** after the payment completed (`completed_at 21:09:31`, `expires_at 21:13:28`). Settlement that has not succeeded inside that window can never succeed. The client-side reconciliation budget is deliberately generous because tenant sweeps have historically lagged, so the two are badly matched: the client is still politely retrying long after the server has closed the door.

## Impact

Money captured, no booking, no automatic recovery, and no operator-facing signal that it happened. Three sessions on one production tenant are currently in this state and had to be found by querying `payment_sessions` for `status = 'paid' AND booking_id IS NULL`. Two of them turned out to have had bookings recreated by hand; one is a real customer with a paid trip and no booking.

## Asks

1. `requirementsFingerprint` should be obtainable while a payment is in flight — either returned on the session read, or `quote` should be permitted (idempotently) for a session whose payment is already settled. Without one of these there is no retry path at all.
2. `idempotency_conflict` must not swallow a corrective edit. Either key the update on the payload rather than the revision, or answer a distinguishable error the client can act on instead of a `200` that looks like success.
3. The payment session's settlement window should outlive a failed commit, or expiry should trigger an automatic reversal so the money does not sit captured against nothing.
4. A tenant-visible signal when a payment settles with no booking. Today this is only discoverable by direct query.

Related: the settlement-time validation asymmetry is filed separately; #4692 is a different cause with the same ending.

### LINKED ISSUE #4734: Session PATCH accepts billing values the commit rejects, so validation fires after the card is captured instead of at the step
## Summary

The Booking Session `PATCH` accepts billing values that the commit schema then rejects. The shopper is therefore told nothing at the step where the field is editable, and is refused at the one moment where a refusal is most expensive — after the card has been captured.

## Reproduction

Runtime `0.82.3`, managed tenant.

1. On the details step, enter a 25-character postal code (`address.postal`).
2. `PATCH /v1/public/catalog/booking-sessions/{id}` → `200`. The value is stored verbatim. Repeated a dozen times, always `200`.
3. Advance and pay by card. Payment captures.
4. The settlement commit answers:

```json
{
  "error": "Too big: expected string to have <=20 characters",
  "code": "invalid_request",
  "details": {
    "issues": [{ "origin": "string", "code": "too_big", "maximum": 20,
                 "inclusive": true, "path": ["contactPostalCode"],
                 "message": "Too big: expected string to have <=20 characters" }],
    "fields": { "formErrors": [],
                "fieldErrors": { "contactPostalCode": ["Too big: expected string to have <=20 characters"] } }
  }
}
```

The write path has no bound; the commit path caps it at 20.

## Why it matters more than a normal validation gap

A commit refusal after capture is not recoverable — see the companion issue: quoting is then refused with `payment_in_flight` and committing requires a `requirementsFingerprint` that only a quote can produce. So a mismatch between the two schemas does not merely produce a bad error message; it produces captured money with no booking and no way back.

`address.postal` is also **not** advertised as a constrained field. The session's `requirements.bookingFields` lists it with no length information at all:

```json
{ "key": "address.line1", "label": "Address line 1", "type": "text", "required": false, "group": "billing" }
```

so a client cannot pre-validate even if it wants to. Note the commit reports the field under a different name (`contactPostalCode`) from the one the write path and the requirements use (`address.postal`), which makes the error hard to map back to an input without a hand-written table.

## Asks

1. Apply the commit's constraints on the write path, so the refusal happens at the step where the field can still be edited.
2. Publish the bounds in `requirements.bookingFields` (`maxLength` and friends) so clients can validate before the shopper leaves the step.
3. Report the field under the name the client supplied it as, or publish the mapping.

Filed alongside the deadlock issue; either fix alone reduces the blast radius, but the pair is what turns a typo into a lost payment.

mihaipxm:
Closes #4734, closes #4733.

## What was wrong

**#4734 — the write path never applied its own widths.** `bookingSelectionPublicV1` has carried `postal: max(20)` since #4298, under a comment promising "an address this schema admits cannot be rejected later by the commit". But `normalizeBookingSelection` **projects** the billing block value by value (`stringValue(...)`) instead of parsing it, so none of those bounds ever ran. The Booking create does parse, which put the only enforcement on the far side of the card capture: a 25-character postal code accepted a dozen times at the step where it could still be edited, and refused once, at settlement, under a field name (`contactPostalCode`) the client had never sent.

**#4733 — the Session's shopping clock kept running under a captured payment.** #4636 stopped the refusal path tearing down the Quote and Hold, but the Session's own TTL was untouched. Two places expire a lapsed Session before asking who is calling — the commit preflight and `consumeCommittedSources` — and expiry releases the Hold, expires the Quote and moves the Session out of `active`: every input a settlement needs. The retry that arrived next answered `session_expired`, and so did every one after it. The window was minutes, not hours, because the payment session's own `expiresAt` is `earliest(session, quote, hold)`.

The two compose into the reported incident: the validation asymmetry supplies a commit refusal, and the expiry makes it permanent.

### #4734

- `BOOKING_SELECTION_BILLING_MAX_LENGTHS` / `..._TRAVELER_...` are one table driving three things that used to hold the number separately: the Zod schema, the Session normalizer, and `requirements.bookingFields[].maxLength`.
- The write path rejects with `invalid_selection` / `value_too_long`, carrying `path` as the caller wrote it (`billing.address.postal`) and `maxLength`. Measured after trimming, because that is the value the commit writes.
- `bookingFieldRequirementV1` and `travelerFieldRequirementV1` publish `maxLength`, and `defaultBookingFields()` now advertises the whole billing address rather than three of six — `address.postal`, `address.line2` and `address.region` were carried by the selection but never published, so a client could not pre-validate the field that bit.
- The selection-to-`contact_*` rename is exported as `BOOKING_SESSION_BILLING_FIELD_TARGETS` and is load-bearing in `billingContact()`, which answers ask 3: the mapping is published rather than reconstructed by hand.
- `bookingSessionRecoveryV1` classifies `value_too_long` as `selectionIncomplete`, not `unknown` — the shopper can fix it, and the generic "the Booking Session rejected this request" is the failure that file exists to prevent.

### #4733

- Money with a processor freezes a Session against expiry, guarded at `expireSession` — the one choke point the commit preflight, the sweep and the lapsed-Session preambles all reach. Only *settled* money holds a Session open: `hasInFlight`'s handoff arm lapses on the Session's own clock, so an abandoned checkout page still gets swept.
- `consumeCommittedSources` and its composite twin no longer consult `expiresAt` for a settlement, for the same reason they already skip the revision check.
- `abandonSession` gets the in-flight guard it was missing. It tears down exactly what a settlement needs, and it is what a "start over" button calls.
- `BookingSessionRecordV1` carries `requirementsFingerprint` (ask 1). Commit requires the fingerprint and only a Quote produced one, which is a dead end precisely while a payment is settled — quoting is refused, correctly. It is a hash of a descriptor the read already returns, so it grants no new authority, and a stale one is still refused by the existing `requirements_changed` comparison.
- A settlement that produces no Booking emits `booking_session.settlement.failed` (ask 4). Emitted before the rethrow, so the handler still fails and still retries; the state was previously discoverable only by querying `payment_sessions` for `status = 'paid' AND booking_id IS NULL`.

### Ask 2 (idempotency) is documented, not changed

The corrective-`PATCH` trap is a client key-derivation defect, and this repo's own client already avoids it: `useBookingSession.updateSelection` keys on `bookingSessionIdempotencyKey(root, "update", revision, bookingSelectionDigest(selection))`. The revision-only key the report describes is not derived here. `updateBookingSessionV1.idempotencyKey` now states the rule and names the derivation, so a client reading the contract does not reinvent the broken one. Server-side, `idempotency_conflict` is already a distinct outcome kind; the `200` is the outcome-envelope convention every rejection uses and is not worth breaking for one arm.

Note that a corrective `PATCH` is still refused while a payment is settled (`payment_in_flight`) — that is #4636's guard and it stays. With the width enforced at the step, the value that triggered it cannot get in.

## Verification

- `packages/catalog` — 835 passed, 7 files skipped
- `packages/commerce` — 188 passed
- `core` 198, `catalog-contracts` 104, `catalog-react` 121, `storefront-sdk` 10, `trips` 220
- `bookings-react` — 248 passed; `admin-extension.test.ts` hit a 5s timeout under parallel-worktree load and passes in isolation (15/15). Unrelated to this diff.
- `biome check .` from root — 6967 files, **0 errors**, 23 warnings, 8 infos
- `packages/catalog` OpenAPI regenerated (`maxLength` on both field descriptors, `requirementsFingerprint` on the Session record). `trips` and `legal` regenerate clean.

The width test is the one worth reading: it drives `defaultBookingFields()` on one side and `bookingCreateSchema` on the other, at the limit and one past it, rather than asserting `20 === 20` against a hand-copied constant — which would pass on the day somebody changes one of them.

## Deploy note

`requirements` gained a field, so every `requirementsFingerprint` changes. In-flight Quotes taken before the deploy are refused with `requirements_changed` / `request_fresh_quote`, which is the designed recovery for exactly this.

### DIFF
--- a/.changeset/booking-session-settlement-widths.md
+++ b/.changeset/booking-session-settlement-widths.md
@@ -3,10 +3,11 @@
 "@voyant-travel/catalog": minor
 "@voyant-travel/catalog-react": patch
 "@voyant-travel/commerce": minor
+"@voyant-travel/core": minor
 ---
 
 Refuse an over-long billing value at the Booking Session write path instead of after the card is captured, and stop a settled payment from being stranded by the Session's own expiry.
 
 The Session's selection normalizer projected the billing block field by field rather than parsing it, so the widths `bookingSelectionPublicV1` declared never ran while the Booking create enforced its own. A 25-character postal code was accepted at every step and refused once, at settlement. `requirements.bookingFields` now publishes `maxLength`, advertises the whole billing address, and the write path rejects with `invalid_selection` / `value_too_long` naming the field as the caller sent it (`billing.address.postal`, not `contactPostalCode`).
 
-A Session whose money is with a processor is no longer expired by the commit preflight or the expiry sweep, and can no longer be abandoned; `BookingSessionRecordV1` carries `requirementsFingerprint`, so a Commit is reachable from a read rather than only from a Quote. A settlement that produces no Booking emits `booking_session.settlement.failed`.
+A Session whose money is with a processor is no longer expired by the commit preflight or the expiry sweep, and can no longer be abandoned; `BookingSessionRecordV1` carries `requirementsFingerprint`, so a Commit is reachable from a read rather than only from a Quote. A settlement that produces no Booking emits `booking_session.settlement.failed`, and `ANALYTICS_FAILURE_REASONS` gains `value_too_long` so the new rejection reaches the breakdown rather than the `unknown` bucket.
--- a/.changeset/booking-session-settlement-widths.md
+++ b/.changeset/booking-session-settlement-widths.md
@@ -1,0 +1,12 @@
+---
+"@voyant-travel/catalog-contracts": minor
+"@voyant-travel/catalog": minor
+"@voyant-travel/catalog-react": patch
+"@voyant-travel/commerce": minor
+---
+
+Refuse an over-long billing value at the Booking Session write path instead of after the card is captured, and stop a settled payment from being stranded by the Session's own expiry.
+
+The Session's selection normalizer projected the billing block field by field rather than parsing it, so the widths `bookingSelectionPublicV1` declared never ran while the Booking create enforced its own. A 25-character postal code was accepted at every step and refused once, at settlement. `requirements.bookingFields` now publishes `maxLength`, advertises the whole billing address, and the write path rejects with `invalid_selection` / `value_too_long` naming the field as the caller sent it (`billing.address.postal`, not `contactPostalCode`).
+
+A Session whose money is with a processor is no longer expired by the commit preflight or the expiry sweep, and can no longer be abandoned; `BookingSessionRecordV1` carries `requirementsFingerprint`, so a Commit is reachable from a read rather than only from a Quote. A settlement that produces no Booking emits `booking_session.settlement.failed`.
--- a/packages/catalog-contracts/src/booking-engine/requirements-contracts.ts
+++ b/packages/catalog-contracts/src/booking-engine/requirements-contracts.ts
@@ -298,6 +298,8 @@ export const travelerFieldRequirementV1 = z.object({
   required: z.boolean(),
   options: z.array(z.object({ value: z.string(), label: z.string() })).optional(),
   appliesToBands: z.array(z.string()).optional(),
+  /** See {@link bookingFieldRequirementV1}'s `maxLength`. */
+  maxLength: z.number().int().positive().optional(),
 })
 export type TravelerFieldRequirementV1 = z.infer<typeof travelerFieldRequirementV1>
 
@@ -308,6 +310,21 @@ export const bookingFieldRequirementV1 = z.object({
   type: z.string(),
   required: z.boolean(),
   group: z.enum(["billing", "company", "preferences"]),
+  /**
+   * The longest value this field accepts, when it is bounded at all.
+   *
+   * A descriptor that says a field exists but not how wide it is leaves a
+   * client unable to pre-validate even when it wants to: `address.postal` was
+   * published with no length information, accepted at 25 characters by the
+   * Session `PATCH`, and refused at 20 by the commit — after the card was
+   * captured (voyant#4734). Publishing the bound is what lets the refusal
+   * happen at the step instead.
+   *
+   * Optional because most fields are genuinely unbounded. Absent means "no
+   * declared limit", never "unknown": the server bounds a field here whenever
+   * it will bound it later.
+   */
+  maxLength: z.number().int().positive().optional(),
 })
 export type BookingFieldRequirementV1 = z.infer<typeof bookingFieldRequirementV1>
 
--- a/packages/catalog-contracts/src/booking-engine/requirements-defaults.ts
+++ b/packages/catalog-contracts/src/booking-engine/requirements-defaults.ts
@@ -11,6 +11,10 @@ import type {
   PaxBandSpecV1,
   TravelerFieldRequirementV1,
 } from "./requirements-contracts.js"
+import {
+  BOOKING_SELECTION_BILLING_MAX_LENGTHS,
+  BOOKING_SELECTION_TRAVELER_MAX_LENGTHS,
+} from "./selection-contracts.js"
 
 /**
  * Default pax bands used when a vertical doesn't supply its own.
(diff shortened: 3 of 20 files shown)


mihaipxm:
Both addressed in 643c2df.

**P1 — settlement authority, not the port probe.** Correct, and the reasoning is the stronger one. `hasInFlight` is optional on `BookingSessionPaymentPorts`, so a runtime that omits it answers `undefined`, the guard reads that as "no money outstanding", and the preflight expires the Session the settlement came to settle. The preflight now keys on `settling` as well — the same fact that already suppresses the revision check and, in this PR, the `expiresAt` check in `consumeCommittedSources`.

Both guards stay rather than one replacing the other: the port probe is what protects the expiry sweep and the shopper-facing paths, which carry no settlement authority at all. New test drives exactly the reported configuration — a payments port with `hasInFlight: undefined` — so the two guards are covered separately rather than one masking the other.

**P2 — core in the changeset.** Added as `minor`. The stated npm-resolution path does not apply (`@voyant-travel/core` is `private: true` and never published), but the conclusion holds for a different reason: core is still changeset-versioned — 0.142.0 came from a changeset — so omitting it leaves the `ANALYTICS_FAILURE_REASONS` widening out of its changelog and skips the dependent cascade.

Full `packages/catalog` suite: 835 passed. Two files needed a rerun for 5s/20s vitest timeouts under parallel-worktree load locally; both pass in isolation and neither touches this diff.

Issue #4706, opened 2026-08-15
Booking contracts can only be issued through the MCP lifecycle command, so an operator cannot issue one from the admin UI


mihaipxm:
## Summary

A booking contract cannot be issued from the admin UI. The direct route refuses, and the only
caller that can satisfy it is the MCP runtime — so on a deployment where the agent is not
wired for contracts, booking contracts accumulate in `draft` permanently and can be neither
issued nor regenerated.

## What happens

`POST /v1/admin/legal/contracts/{id}/issue` on a booking-linked contract returns **400**:

> Managed booking contract revisions must be issued through the reviewed lifecycle command.

The admin UI renders its `Issue` button unconditionally on the contract detail page, so the
only affordance an operator can see is one that always fails.

## The gate is already there — one caller just knows the password

`packages/legal/src/contract-lifecycle-command.ts:238`:

```ts
const issued = await issueContract(db, contract, { allowManagedReviewWorkflow: true })
```

So this is not an architectural wall. `issueContract` already accepts an opt-in, and
`executeLegalContractLifecycleCommand` passes it. The direct admin route does not, and there is
no supported way for it to.

`executeLegalContractLifecycleCommand` is invoked **only** from
`packages/legal/src/mcp-runtime.ts` (~321-351). It has no admin REST route — I probed
`/regenerate-document`, `/regenerate`, `/documents`, `/generate-document`, `/document` and
`/attachments/generate`; all 404. (`/render` exists but resolves against templates, not
contract instances.)

Net effect: the reviewed lifecycle is reachable only by an agent. An operator with a browser
cannot issue a contract at all.

## Why this is the wrong shape

Contracts are customer-facing legal documents. Making them issuable *only* through an AI agent
means a deployment that doesn't wire the agent for contracts — or an operator who simply wants
to click a button — has no path. Issuing a contract should not require an agent to be present,
configured and authorized.

There is a host-side counterpart to this (a managed key needs a scope before its agent can
reach the contract tools), filed separately in the host repo. But fixing only that would leave
the capability agent-only, which is why this one is the primary fix.

## Impact

Observed on a deployment where every booking contract generated since migrating to the managed
runtime sits in `draft`; contracts predating it are `issued`, so this is a regression rather
than a capability that never existed. Compounding it, a draft's PDF is rendered once at
generation time and there is no regenerate path, so a draft whose content is wrong can be
neither corrected nor re-issued — only voided or deleted.

## Asks

[...]

PR #4748, opened 2026-08-16
Show flight availability where the date is chosen

mihaipxm:
The flights date picker rendered an undecorated calendar — every day looked identical, so nothing on screen said which dates the operator's providers actually fly or what they cost. Nothing behind it could have said so either: there was no fare-calendar concept anywhere in the flights stack.

## What changed

**`flight/fare-calendar`** — new capability + `searchFareCalendar` on `FlightConnectorAdapter`, served as `POST /v1/admin/flights/fare-calendar` (92-day window cap). The picker shows the cheapest indicative price per day, bands it cheap/mid/expensive against the visible window, and strikes out days with no service. Each picker prices the leg it selects, so the return shows the fare for coming back rather than repeating the outbound.

The demo adapter prices every day by running the *real* offer synthesis rather than a parallel price model, so a number the picker shows and a search on that day can never disagree — a test pins exactly that. Demo supply also gained a weekly operating pattern on thin long-haul routes, which is what gives the calendar real gaps instead of a uniform wall of availability.

**`flight/served-markets`** — new capability + `listServedMarkets`, served as `GET /v1/admin/flights/served-markets`. The airport picker leads with the operator's own network. **It ranks, it never filters**: a stale declaration must not be able to hide an airport, so the remainder group always carries everything else.

**Operator signal** — the airport picker also groups by routes actually searched, remembered per browser (selection caches display fields; only a submitted search counts as a use). And `GET /reference/airports` is now deterministically ordered — previously the "first 200" a picker showed was whatever Postgres happened to hand back.

**Airline visibility** — offer rows name the airline and its flight numbers. `carrierName` was already resolved and passed down; it was only being spent on the logo's `alt` text.

**`packages/ui`** — `Calendar`/`DatePicker` gain `dayAnnotation` for a secondary line under a day's number. The shadcn `CalendarDayButton` was already a flex-column with `[&>span]` styling, so this fills a seam that was already there.

## Degradation

A connector declaring neither capability answers 501 and every surface falls back to exactly what it did before — a plain working calendar, an ungrouped airport list. That is the common case, not a failure, and it's covered by tests.

## New checks

Both new checks are declarative additions to existing capability/route machinery rather than new `scripts/check-*.mjs` — no new links in the `verify:architecture` chain.

## Verification

- `flights-contracts`, `flights`, `flights-react`, `ui`, `flights-demo-api`: build, typecheck, lint, test all green
- 18 new component tests in `flights-react` (first jsdom tests in that package; `@testing-library/react` + `jsdom` added as devDeps, matching `media-react`)
- `i18n:check` green, en+ro parity for every new string
- `verify:openapi-drift`, `route-conformance`, `route-authoring`, `graph-conformance`, `symbol-policy`, `public-surface`, `boundary`, `client-package-boundaries`, `typecheck-coverage`, `vitest-include`, `agent-tool-coverage`, `agent-write-coverage`, `dep-paths`, `dist-configs`, `flights-runtime-authority` — all green

Not verified in a running browser: the operator's flights adapter is deployment-wired outside this repo, so there is no local path that serves a fare calendar. Behaviour is pinned in a real DOM instead — prices rendered, band classes on the right cells, sold-out days unselectable, unquoted days still selectable, 501 degradation.

### DIFF
--- a/.changeset/flights-fare-calendar-and-airport-scoping.md
+++ b/.changeset/flights-fare-calendar-and-airport-scoping.md
@@ -1,0 +1,32 @@
+---
+"@voyant-travel/flights-contracts": minor
+"@voyant-travel/flights": minor
+"@voyant-travel/flights-react": minor
+"@voyant-travel/ui": minor
+---
+
+Show flight availability where the traveller chooses it.
+
+Adds two capability-gated connector methods and the UI that consumes them:
+
+- `flight/fare-calendar` (`searchFareCalendar`) quotes a window of departure
+  dates, so the date picker shows the cheapest indicative price per day, bands
+  it cheap / mid / expensive against the visible window, and strikes out days
+  the provider doesn't fly. Served as `POST /v1/admin/flights/fare-calendar`,
+  capped at a 92-day window.
+- `flight/served-markets` (`listServedMarkets`) declares the airports a
+  connection sells, so the airport picker leads with the operator's own
+  network. It ranks, it never filters — every airport stays reachable.
+
+The airport picker also groups by routes this operator has actually searched,
+remembered per browser, and the airport reference list is now deterministically
+ordered instead of returning an arbitrary slice.
+
+Flight offer rows now name the airline and its flight numbers instead of
+relying on the carrier logo alone.
+
+`Calendar` and `DatePicker` gain a `dayAnnotation` prop for rendering a
+secondary line under a day's number.
+
+Connectors that declare neither capability answer 501 and every surface
+degrades to its previous behaviour.
--- a/apps/flights-demo-api/src/routes.ts
+++ b/apps/flights-demo-api/src/routes.ts
@@ -4,6 +4,8 @@
  * integration source. Each endpoint maps 1:1 to an adapter call.
  *
  *   POST   /search                       searchFlights
+ *   POST   /fare-calendar                searchFareCalendar
+ *   GET    /served-markets                listServedMarkets
  *   POST   /price                        priceOffer
  *   POST   /book                         bookFlight
  *   GET    /orders                       listOrders
@@ -17,6 +19,7 @@
 
 import type {
   AncillaryRequest,
+  FareCalendarRequest,
   FlightBookRequest,
   FlightOrder,
   FlightOrderStatus,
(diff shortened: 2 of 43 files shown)


mihaipxm:
Addressed the two real findings; the P1 is a false positive.

**P1 — model authorship: not reproducible.** Both commits are authored *and* committed by `Mihai Ungureanu <mihai@pixelmakers.com>`, locally and as GitHub records them:

```
$ gh api repos/voyant-travel/voyant/commits/37ba41edf4 --jq '{author: .commit.author, committer: .commit.committer}'
{"author":{"email":"mihai@pixelmakers.com","name":"Mihai Ungureanu"},
 "committer":{"email":"mihai@pixelmakers.com","name":"Mihai Ungureanu"}}
```

No `Co-authored-by` trailer, no generated-with footer, nothing prohibited in any branch, commit, or PR text. Nothing to recreate.

**P2 — stale published OpenAPI document: correct, and worse than described.** `packages/flights/openapi/admin/flights.json` says "Do not edit by hand" but nothing regenerated it, *and* it was not in `scripts/checks/openapi/generated-specs.json` — which is why `verify:openapi-drift` passed green while the contract was stale. That's the exact failure that file's own `$comment` warns about, and the one legal hit in #4706.

Fixed the same way: `packages/flights/scripts/generate-openapi.ts` drives the real route module through `stampModuleMetadata`, registered with the drift checker (5 → 6 documents). Regenerating reproduces the existing twelve operations **byte for byte** and adds only the two new paths — which is the evidence that the generator takes the path the artifact originally came from, rather than a plausible-looking substitute.

One thing the suggestion would have missed: the artifact carries `x-voyant-unit-id` and `x-voyant-package-name`, which the composed operator app stamps and `stampModuleMetadata` does **not**. A naive regeneration would have silently stripped both fields off all twelve existing operations. The generator re-applies them. The generator also replaces everything under `/v1/admin/flights` rather than merging, so a retired route actually disappears instead of outliving itself.

**P2 — impossible dates: correct.** Confirmed both failure modes:

```
2026-02-31 -> Date.parse -> 2026-03-03   (silently rolls forward; connector quotes the wrong month)
2026-13-01 -> NaN -> calendarWindowDays returns 0 -> slips under the 92-day cap
```

`isoDateSchema` now round-trips the parsed year/month/day against the input. Covered by tests for `2026-02-31`, `2026-13-01`, `2026-04-31`, `2026-00-10`, plus a leap-day pair (`2028-02-29` accepted, `2027-02-29` rejected) so the refinement can't pass by rejecting everything.

All touched packages green: contracts 3, flights 150, flights-react 27, flights-demo-api 13; typecheck, lint, and `verify:openapi-drift` clean.

Issue #4627, opened 2026-08-14
Move the parts of packages/storefront that have real domain homes; keep the composition layer


mihaipxm:
`packages/storefront` (→ `packages/public-api`, #4624) depends on **eleven** domain modules — auth, bookings, catalog, commerce, cruises, finance, flights, identity, legal, payments, relationships — and its migrations create exactly **one** table, `storefront_verification_challenges`. Almost everything in it is composition over other modules' state.

That looks like a package that should be dissolved. It mostly isn't, and the reason is structural.

## Why it cannot be fully dissolved

The modules underneath it depend on each other: `commerce → catalog`, `catalog → bookings`, `trips → catalog, commerce, bookings`. A read model that needs two of them has nowhere below to live.

Concretely: the departures projection needs catalog data *and* commerce pricing/promotions. Move it into `catalog` and you get `catalog → commerce → catalog`. A cycle. It has to sit above both, which is where it already is.

So this is a **partial** extraction: move what has a genuine home, keep the composition layer.

## Move out

- [ ] `payment-link/` → `packages/payments` — already owns `checkout-handoff` and the provider catalog
- [ ] `transport-eligibility` (`service-transport-eligibility.ts`, `validation-transport-eligibility.ts`) → `packages/flights`
- [ ] `guest-booking-guard.ts` → `packages/bookings`
- [ ] `verification/` and the `storefront_verification_challenges` table → `packages/identity` — the one piece of owned domain state in the package
- [ ] `shopping/` **trip-selections** — the `/trip-selections` and `/trip-selections/book` routes, `StorefrontTripSelectionCreate`, `storefrontTripSelectionsRuntimePort` → `packages/trips`, which already owns checkout and the booking-session composite handler and already mounts its own `/v1/public` routes (proposals, bookings)

## Keep

Cross-module read models with no cycle-free home below:

- `departures-read-model.ts` — the KV doc cache, TTL, key scheme and invalidation subscriber
- `service-departures*` — core, offers, pricing, pricing-context, price-preview, extensions
- `shopping/` — **narrowed to `/search`**: cross-vertical query over `products | accommodations | cruises | charters`, market/locale/currency scoping, opaque reference issuance, the closed trust-plane seam
- offers resolution

These are the reason the layer exists. Naming it `public-api` (#4624) is what makes that legible.

[...]

PR #4759, opened 2026-08-16
Give ships a surface of their own (recovery of #4722)

mihaipxm:
Replays #4722 onto `main`. That PR merged, but into `fix/catalog-detail-missing-info` rather than `main`, and its base had already gone to main via #4718 nineteen minutes earlier. The squash landed on a branch nothing merges from, so the Ships surface never reached `main` — `packages/cruises-react/src/admin/` does not exist there, and the changeset never ran, so none of the three packages were bumped or published.

Same two commits, cherry-picked onto current `main`. Only the `cruises-react` manifest needed resolving: `@voyant-travel/storefront-react` is gone, replaced by `@voyant-travel/public-api-react` (#4627/#4649), so the peer, peer-meta and dev entries take main's name and add the surface's genuinely new peers, `@voyant-travel/admin` and `@voyant-travel/distribution-react`. The lockfile was regenerated from main's rather than merged by hand.

`verify:typecheck-coverage` needs no baseline edit: main already carries 55 entries with `cruises-react` absent, and the checker reports no new unchecked package and no tightening available.

## Verification

- `pnpm install --frozen-lockfile` — clean, so the regenerated lockfile matches the manifests
- `build` across the `cruises` + `cruises-react` dependency closure — exit 0, zero `error TS`
- `typecheck` on `cruises`, `cruises-react`, `i18n` — exit 0 (covers `tests/**` via the #4243 tsconfig split)
- `test` on the same three — 318 passed, and the three new files were confirmed to run by name rather than by the total: `tests/unit/ships-search.test.ts` (3), `src/admin/ship-presentation.test.ts` (9), `tests/admin-destinations.test.ts` (3)
- `pnpm verify:architecture` — the full chain, exit 0
- `biome check` on the touched packages — exit 0 (the 2 warnings are pre-existing in untouched `cruises/src`)

Closes the gap left by #4722.


mihaipxm:
Good catch on the destination bindings — confirmed and fixed in d82c2e2.

Both ship routes now declare the key they satisfy (`cruiseShip.list` / `cruiseShip.detail`); the detail route already names its param `shipId`, so no `destinationParams` mapping is needed.

You were also right that the type-only fixture was the wrong guard, and it is worth recording *how* wrong: it declared a resolver map inline and asserted on its own literal, never touching the extension. I checked by removing both bindings again — the three new assertions fail, and that original type-level test still passes. It now drives `buildAdminExtensionDestinations` over the real extension, which is the collector the host actually runs, and covers a host-renamed mount path too.

Issue #4626, opened 2026-08-14
SDK surface: split clients by API surface — @voyant-travel/public-api and @voyant-travel/admin-api


mihaipxm:
Split the client SDKs by **API surface**, not by where the code runs, and name them after the surface they call.

## Why not "frontend SDK" vs "admin SDK"

Both halves get used to build front-ends — the admin API exists partly so back-office UIs can be built faster. And the public API client runs server-side constantly: SSR, a route handler, a BFF, a mobile backend. "Frontend" is ambiguous in both directions.

The surfaces are stable and the credential is orthogonal to them (see #4625: two surfaces, two credentials, SK crosses both).

| Package | Surface | Credential | Runs |
|---|---|---|---|
| `@voyant-travel/public-api-client` | `/v1/public/*` | PK **or** SK | browser + server |
| `@voyant-travel/admin-api-client` | `/v1/admin/*` | SK only | server only |

## Naming

**`-client`, not `-sdk`.** Two reasons.

`-sdk` is actively ambiguous here: `@voyant-travel/admin-sdk` next to the existing `@voyant-travel/admin-extension-sdk` — a completely different thing, for building framed admin extensions — would be confused constantly. `admin-api-client` vs `admin-extension-sdk` reads as two different things at a glance.

And the bare surface names are taken by the server side. `packages/storefront` becomes `packages/public-api` (#4624) because that package really is the layer that composes the public surface across domains — it depends on eleven domain modules and owns one table. So the client needs a suffix, and `-client` is the honest one: it is a client *for* the public API, not the public API itself.

### 1. `@voyant-travel/public-api-client`

- [ ] Rename from `packages/storefront-sdk`
- [ ] Isomorphic — the same client in a browser bundle and on a server
- [ ] Credential is a constructor concern: `createPublicApiClient({ publishableKey })` vs `({ secretKey })`
- [ ] `packages/storefront-react` → `packages/public-api-react` follows (tracked in #4624)

### 2. `@voyant-travel/admin-api-client`

- [ ] New package; does not exist today (`packages/admin*` are the admin application, not a client)
- [ ] SK only — refuse to construct with a PK, and fail at construction rather than at first request
- [ ] Never bundled for a browser

### 3. Type the credential split

Once #4625 lands `x-voyant-key-kind` on every route, generation can encode it: **a client constructed with a PK only exposes PK-callable operations.** SK-only operations do not exist on the type.

- [ ] Generate both clients with the key-kind split reflected in the types
[...]

PR #4802, opened 2026-08-17
Generate the Admin API client, and give the media document an openapi version

mihaipxm:
Completes **#4626**. `@voyant-travel/admin-api-client` is typed from all 60 admin OpenAPI documents, one module per document:

```ts
import type { paths } from "@voyant-travel/admin-api-client/finance"
const finance = createAdminApiClient<paths>({ secretKey, baseUrl })
```

One package per surface with per-module subpaths — not a flat client with several hundred operations, and not 60 packages. Delivered through a wildcard export (precedent: `auth-react`'s `./components/*`).

## It stays private, on evidence

I flagged "publish or not" as an open decision earlier. Checking it, it isn't one:

- **All 31 `*-react` packages — every in-repo consumer of `/v1/admin/*` — are `private: true`.** Zero publishable.
- The published surface is **15 packages**: eight `*-contracts`, `schema-kit`, `ui`, `payments`, `app-manifest`, `graph-contracts`, `admin-extension-sdk`, `public-api-contracts`. **None is an API client.**
- An outside integrator holds a secret key and speaks HTTP. What they need is the **OpenAPI document**, which already ships in the tarball and is rendered by the in-app reference — they can generate a client on their schedule instead of being coupled to our release train.

#4256's publish argument (*"an SDK external storefront developers cannot install is of limited use"*) is about the **public** client. It doesn't transfer to admin. Publishing this would make it the first API client on npm and reopen the #4098 closure for no consumer that exists.

## The credential is refused at construction

#4626 §2 asks for SK-only, failing at construction. `createAdminApiClient` throws `AdminApiClientCredentialError` for a publishable key, an unrecognised prefix, or an empty value — a `vpk_` is browser-resident, and if one reaches this constructor someone is about to ship it where it can be read.

The kind comes from `classifyPublicApiKeyToken` in `@voyant-travel/core`, **imported, not re-implemented**. That module states the reason next to itself: two copies of the prefix table drifting is an auth bypass. The tests derive fixtures from `PUBLIC_API_KEY_PREFIXES` rather than writing `vsk_`, so a test cannot keep passing while the real table moves.

**No publishable admin variant**, deliberately: every `/v1/admin/*` route is secret-only by construction, so a publishable subset would be an empty type pretending to be a boundary. The PK/SK split belongs to the public surface, where `public-api-client/generated` carries it.

## A document that was not an OpenAPI document

Generation failed on document 41 of 60:

```
Error: Unsupported schema format, expected `openapi: 3.x`
  packages/media/openapi/admin/media-library.json
```

**It had no `openapi` field at all.** `media`'s generator calls `getOpenAPI31Document({ info, servers })` and never passes `openapi` — `legal`'s passes `openapi: artifact.openapi`, and the builder emits only what it is given.

Why nothing caught it: `verify:openapi-drift` **regenerates and compares**, and the generator omitted the field exactly as consistently as the artifact lacked it. Both sides were wrong in the same way, so they agreed. A round-trip check cannot see a fault its own generator reproduces — worth remembering, because that is the shape of the check we are adding everywhere.

Survey of all 82 documents: 80 are `3.1`, this one was fieldless (now `3.1.0`), and `navigation-preferences` is `3.0.0` — harmless, not touched here.

## Verification

- `pnpm verify:architecture` — full chain, exit 0, tree unchanged throughout
- `verify:openapi-drift` — **78 generated files**, up from 18
- typecheck **1.25s at 417 MB** for 3.4 MiB across 60 modules; proven non-vacuous by injecting an error into `team.ts` (exit 2, 1 error) and clearing it
- build emits 60 `.d.ts`; `verify:dep-paths`, `verify:dist-configs`, `verify:package-exports` all pass
- 7 tests. The security-relevant one is proven non-vacuous: reordering the header spread so a caller could override `x-api-key` makes it fail

No changeset — both `media` and `admin-api-client` are private, and a changeset naming a private package breaks `release:plan`.

## Not in this PR

`vsk_`/`vpk_` are not pinned by `symbol-policy.json`, and this adds a consumer. 18 non-test files contain the literals, though many are likely doc comments that `literalsOnlyIn` ignores. A real candidate for a rule, sized separately.

**#4256 still needs P0** (16 of 82 documents drift-registered; 24 packages with no generator) **and P3** (40 hand-rolled `client.ts` files, 4,210 lines). P0 should land before P3: a drifted document produces a drifted client, and that only starts to hurt when consumers depend on it.


mihaipxm:
All three review comments were valid and are fixed in f4c9174.

**1. Reinstate the secret key after request headers** — confirmed against the real client before fixing: `client.GET(path, { headers: { "x-api-key": vpk_… } })` sent `vpk_leaked0000`. The constructor check could be walked straight past, and my tests missed it because they only ever set headers on the *constructor* — the case that already worked. A request middleware now sets the credential after openapi-fetch's merge. The new test fails without the middleware.

**2. Validate the dynamically generated output set** — confirmed by planting an admin document: it produced an unregistered module and `verify:openapi-drift` passed anyway. The generator now prunes a module whose document is gone, and refuses to finish if it wrote a file `generated-specs.json` does not list. Both verified — adding a document exits 1 naming the file; removing one logs `removed …/legal.ts (its document is gone)`.

That fix exposed a second-order bug: the prune deleted the hand-written barrel in `src/generated/`. Anything inside a generated directory has to be generated, or the prune cannot distinguish stale output from deliberate work — so the barrel is generated now too, and registered.

**3. Keep test sources out of the build artifact** — correct, and worse than scoped here. The same pattern meant `@voyant-travel/public-api-contracts`, which **is published**, shipped `request-contracts.test.js` and its declarations in the 0.1.0 and 0.2.0 tarballs. Both builds now exclude `*.test.ts`, with a changeset for the published package. Typechecking is unchanged — verified by injecting a type error into each package's tests and watching both fail.

Full `verify:architecture` green; drift now covers 79 files.

PR #4814, opened 2026-08-17
Build the operator webhooks document with the 3.1 builder it claims

mihaipxm:
A one-word bug with a measurable consequence.

```ts
const document = app.getOpenAPIDocument({   // ← the 3.0 builder
  openapi: "3.1.0",                          // ← a label, not an instruction
```

`getOpenAPIDocument` builds a **3.0** document. Passing `openapi: "3.1.0"` only relabels it. So every `z.string().nullable()` published as 3.0's `nullable: true` — a keyword **3.1 does not have and silently ignores**.

## What it cost

Regenerating the admin client against the corrected document:

| | before | after |
|---|---|---|
| fields typed `string \| null` | 17 | **28** |

Eleven fields the API can return as null were typed as never null. A consumer repointed onto those types would be told, with full type-checker confidence, that a value is always present. This is the concrete form of the argument for doing #4256 P0 before P3.

## Why nothing caught it

- **`verify:openapi-drift` can't**: it regenerates and compares. The generator produced `nullable` exactly as consistently as the artifact contained it, so both sides were wrong in the same way and agreed. A round-trip check proves reproducibility, not correctness.
- **A 3.1 parser can't**: `openapi-typescript` accepts the old document, because `nullable` is simply an unknown keyword to it, not an error. It parsed fine and produced wrong types.
- **The version string can't**: it was declared, not derived.

## Scope of the change, verified

Only nullability moved:

- paths identical (10), no top-level field changed except 3.1's empty `webhooks: {}`
- `nullable` 33 → 0; `type` +11
- every remaining difference is an `anyOf` branch that was a bare `{nullable: true}` becoming `{type: "null"}` — the 3.1 spelling of the same thing

I diffed the two documents with 3.0 `nullable` normalised into the 3.1 union to confirm nothing else changed, rather than reading the raw diff and assuming.

## Verification

- `pnpm verify:architecture` — full chain, exit 0
- `verify:openapi-drift` green after regenerating the client (it correctly went red first — the document and its generated module are wired together)
- `openapi-typescript` parses the result; `webhook-delivery` 57 tests pass

No changeset: `webhook-delivery` and `admin-api-client` are both private.

## Follow-on

This is the only generator using the 3.0 builder; the other six already call `getOpenAPI31Document`. It clears **33 of the 87** `nullable` keywords across the corpus, leaving 54 in four documents — `setup` (34), `notifications` (15), `finance` (3), `identity` (2). Three are hand-written. `finance`'s three sit in regions its generator does not own, which is worth understanding before #4256 P0 adds 24 more generators using the same merge pattern.

### DIFF
--- a/packages/admin-api-client/src/generated/operator-webhooks.ts
+++ b/packages/admin-api-client/src/generated/operator-webhooks.ts
@@ -252,7 +252,7 @@ export interface operations {
               description: string | null
               createdAt: string
               updatedAt: string
-              lastDeliveryAt: string | unknown
+              lastDeliveryAt: string | null
               failureCount: number
             }[]
           }
@@ -322,7 +322,7 @@ export interface operations {
                 description: string | null
                 createdAt: string
                 updatedAt: string
-                lastDeliveryAt: string | unknown
+                lastDeliveryAt: string | null
                 failureCount: number
               }
               secret: string
@@ -403,7 +403,7 @@ export interface operations {
               description: string | null
               createdAt: string
               updatedAt: string
-              lastDeliveryAt: string | unknown
+              lastDeliveryAt: string | null
               failureCount: number
             }
           }
@@ -536,7 +536,7 @@ export interface operations {
               description: string | null
               createdAt: string
               updatedAt: string
-              lastDeliveryAt: string | unknown
+              lastDeliveryAt: string | null
               failureCount: number
             }
           }
@@ -615,7 +615,7 @@ export interface operations {
               description: string | null
               createdAt: string
               updatedAt: string
-              lastDeliveryAt: string | unknown
+              lastDeliveryAt: string | null
               failureCount: number
             }
           }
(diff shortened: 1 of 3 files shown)


mihaipxm:
**Correction.** This PR (and #4817) said that OpenAPI 3.1 ignores the 3.0 `nullable` keyword, so a client generated from such a document would be typed to expect a value the API may not send. That is wrong, and I repeated it in several places.

A controlled fixture, same schema in both spellings:

```
with 3.0 `nullable`:  a: string | null;  b: string[] | null;  c: {…} | null
with 3.1 union:       a: string | null;  b: string[] | null;  c: {…} | null
```

`openapi-typescript` honours both identically. Converting three other documents to the 3.1 spelling changed the generated client types by nothing at all.

**What is still true about this PR:** the operator webhooks client did move from 21 to 32 `| null`, and that improvement is real. The cause was switching `getOpenAPIDocument` (the 3.0 builder) to `getOpenAPI31Document`, which regenerated the document from the live routes and fixed genuine under-documentation — the same class of drift later found in finance and legal. The measurement was right; my explanation of the mechanism was not.

The reason recorded in `check-openapi-dialect.mjs` and `nullable-baseline.json` is corrected in #4823: the dialect rule holds **conformance**, not correctness — a document declaring 3.1 should be readable by anything implementing 3.1, rather than only by tools that still accept the older keyword.

PR #4817, opened 2026-08-17
Hold every OpenAPI document to the 3.1 dialect it claims

mihaipxm:
`navigation-preferences` was the last document on `3.0.0`. Converting it is small — six `nullable` keywords, all on **untyped** schemas, where 3.0's "any type, may be null" and 3.1's bare `{}` mean the same thing, so the keyword is simply dropped. Validated by parsing the result with a real 3.1 parser rather than by reading it.

The checker is the point of the PR.

## Three defect shapes, none of them visible before

| shape | example | why nothing saw it |
|---|---|---|
| no `openapi` field at all | `media-library.json` | its generator omitted the field as consistently as the artifact lacked it |
| declares 3.0 while everything else is 3.1 | `navigation-preferences.json` | nothing compared documents to each other |
| declares 3.1 and uses `nullable` anyway | 5 documents, 87 keywords | 3.1 parsers treat `nullable` as an unknown keyword, not an error |

The first two are why this exists. `verify:openapi-drift` **cannot** catch either, and the reason generalises: it regenerates and compares, so a fault its own generator reproduces faithfully is invisible to it forever. It proves reproducibility, not correctness. That property is about to matter a great deal more, because #4256 P0 adds 24 more generators.

## The third shape is the dangerous one

3.1 has no `nullable`. `{"type": "string", "nullable": true}` reads as a plain non-nullable string to every 3.1 tool — **including the client generators**. #4814 fixed the largest instance at its source and measured the cost: regenerating the admin client moved fields typed `string | null` from **17 to 28**. Eleven fields the API can return as null were typed as never null.

## Ratcheted, not fixed

54 keywords remain in four documents — `setup` (34), `notifications` (15), `finance` (3), `identity` (2). Counts **may only shrink**; a document not in the baseline may never gain one; and when a count falls the check fails and tells you to lower the baseline in the same commit.

That last rule already earned itself: after rebasing on #4814 this branch went red with `operator-webhooks.json: 33 -> 0` and I lowered the baseline, rather than the improvement going unrecorded.

Three of the four are hand-written. `finance`'s three sit in a region its generator does not own — worth understanding before more generators adopt the same merge shape.

## Verification

- `pnpm verify:architecture` — full chain, exit 0, tree unchanged throughout, zero failure markers
- checker proven in **four** directions, each by injecting the fault: missing `openapi` field; a document reverting to 3.0.0; a clean document gaining a `nullable`; and a baselined count falling without the baseline being lowered
- `verify:openapi-drift` 79 documents; `check-openapi-dialect` 82 documents

Deliberately imperative rather than a rule file: the assertion is about the *contents* of 82 JSON documents, which no existing rule file models.


mihaipxm:
**Correction.** This PR said 3.1 ignores the 3.0 `nullable` keyword, so `{type: "string", nullable: true}` would read as a plain non-nullable string to every 3.1 tool "including the client generators". That is wrong.

A controlled fixture, same schema in both spellings, shows `openapi-typescript` honours both identically:

```
with 3.0 `nullable`:  a: string | null;  b: string[] | null;  c: {…} | null
with 3.1 union:       a: string | null;  b: string[] | null;  c: {…} | null
```

Converting `setup`, `notifications` and `identity` to the 3.1 spelling changed the generated client types by nothing — 16 `string | null` in `setup.ts` before and after.

**What is still true:** `nullable` is genuinely not an OpenAPI 3.1 keyword, so a document declaring 3.1 and using it is non-conformant. The ratchet is worth keeping for that reason. But the consequence I claimed — wrong types reaching consumers — does not hold for this toolchain, and the ratchet should not have been justified that way.

`check-openapi-dialect.mjs` and `nullable-baseline.json` are corrected in #4823 to record conformance as the reason.

PR #4834, opened 2026-08-17
Generate six more admin documents from their routes

mihaipxm:
Six documents across five packages, plus one route fix the exercise produced.

| document | paths | outcome |
|---|---|---|
| `charters` admin | 13 | **byte-identical** |
| `action-ledger` | 10 | **byte-identical** |
| `charters` public | 7 | `x-voyant-api-id` added on every operation |
| `custom-fields` | 6 | **9 responses, 9 operationIds, 7 parameter sets absent** from the document |
| `navigation-preferences` | 3 | metadata added; schemas already correct |
| `event-catalog` | 1 | additive, plus two equivalent spellings |
| `action-ledger-health` | 2 | schemas added; **the routes were corrected**, see below |

Four regenerate byte-identical or schema-clean. That is worth stating: the generators are not rewriting documents into their own idiom, so a diff means something when one appears.

## The document was more accurate than the code

`/health` and `/health/check` shared one `healthResponses` object, so the routes described the canary endpoint with the projection wording — *"The action ledger and domain projections are healthy"* — while the hand-written document said what it actually reports: *"The canary and all drift checks passed"*.

Generating from the routes would have replaced correct prose with generic prose. Instead `/health/check` now has its own response definitions carrying the document's wording, so **both** are right. First time in this series the exercise improved runtime code rather than an artifact.

## Two template corrections

**Path convention.** Route modules use both: some mount at `/` with prefix-relative paths, others declare absolute paths. Prefixing unconditionally produced `/v1/admin/navigation-preferences/v1/admin/navigation-preferences` — a diff that reads as a wholesale path rename rather than the mistake it is. The template now normalises.

**Equivalent spellings.** `event-catalog`'s `const: "voyant.event-catalog.v1"` became `{type: "string", enum: [...]}` and `additionalProperties: true` became `{}`. Checked rather than assumed — same constraints, zod-openapi's spelling.

## Verification

- `pnpm verify:architecture` — full chain, exit 0, zero failure markers, tree settled throughout
- `verify:openapi-path-ownership` **658/658**; `verify:openapi-drift` **89 documents**; dialect 82 documents, 0 `nullable`
- action-ledger 244 tests, custom-fields 80, event-catalog 4, navigation-preferences 13

Part of #4256 P0.

### DIFF
--- a/packages/action-ledger/openapi/admin/action-ledger-health.json
+++ b/packages/action-ledger/openapi/admin/action-ledger-health.json
@@ -7,29 +7,305 @@
   "paths": {
     "/v1/admin/action-ledger/health": {
       "get": {
-        "summary": "Inspect action ledger drift and health",
         "x-voyant-api-id": "@voyant-travel/action-ledger#health-extension.api",
         "parameters": [
           {
+            "schema": {
+              "type": "string",
+              "format": "date-time"
+            },
+            "required": false,
             "name": "createdAtFrom",
-            "in": "query",
-            "schema": { "type": "string", "format": "date-time" }
+            "in": "query"
           },
           {
+            "schema": {
+              "type": "integer",
+              "minimum": 1,
+              "maximum": 100
+            },
+            "required": false,
             "name": "sampleLimit",
-            "in": "query",
-            "schema": { "type": "integer", "minimum": 1 }
+            "in": "query"
           }
         ],
         "responses": {
-          "200": { "description": "The ledger and its domain projections are healthy" },
-          "503": { "description": "A ledger or domain projection health check failed" }
-        }
+          "200": {
+            "description": "The action ledger and domain projections are healthy",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "data": {
+                      "type": "object",
+                      "properties": {
+                        "ok": {
+                          "type": "boolean"
+                        },
+                        "canary": {
+                          "type": ["object", "null"],
+                          "properties": {
+                            "ok": {
+                              "type": "boolean"
+                            },
+                            "actionId": {
+                              "type": "string"
+                            },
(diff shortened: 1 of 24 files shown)


mihaipxm:
Good catch — this was a real design flaw and it is fixed in `feat/openapi-generators-batch-6` (ea64f49, 579191c), which supersedes the copy logic in every generator added here.

**The defect.** The copy was unconditional, so a route that changed its own `operationId`, `summary`, `tags` or `x-voyant-api-id` would have had the change overwritten by the committed artifact — and `verify:openapi-drift` would then reproduce the stale contract indefinitely. A check meant to detect drift would have been guaranteeing it, in the dimension it looked most confident about.

**Why it was written that way.** `stampModuleMetadata` derives placeholder summaries for operations that declare none — `"GET /v1/admin/apps"` where a human wrote `"List app registrations"`. A naive live-wins rule replaced 24 curated summaries in `apps` with placeholders, which is how the blanket copy got in. So the fix is a three-way precedence rather than a reversal:

> **what the route declared → what the document already said → what stamping derived**

implemented by snapshotting operation metadata *before* stamping fills in defaults, so "the route said something" is distinguishable from "stamping invented something".

**Verified both directions.** All 116 documents regenerate byte-identical, so the change is behaviour-neutral today — it only bites when a route actually changes its metadata. And changing a route's `summary` now propagates to the document (`"CHANGED BY THE ROUTE"`), which the previous logic silently reverted.

One generator needed a follow-up: `public-document-delivery` omits `stampModuleMetadata` so it need not depend on `@voyant-travel/hono`, and the bulk patch anchored on that call skipped it — it got the consumer without the declaration. Fixed, and every generator audited for the same gap rather than trusting the patch count.

Full chain green: `verify:openapi-drift` 116 documents, `verify:openapi-path-ownership` 1014/1014.

Issue #4837, opened 2026-08-17
RFC: First-class inquiry management and sales conversion


mihaipxm:
# RFC: First-class inquiry management and sales conversion

- **Status:** Draft (2026-08-18)
- **Owners:** relationships + proposals + bookings + storefront + admin
- **Related:**
  [ADR-0018](../adr/0018-proposals-as-travel-native-bespoke-sales-artifact.md),
  [ADR-0019](../adr/0019-booking-v1-commitment-point-policies.md),
  [storefront architecture](./storefront-architecture.md),
  [packaged admin RFC](./packaged-admin-rfc.md), and
  [agent tool library](./agent-tool-library.md)

This RFC gives travel agencies one durable place to receive, triage, work, and
resolve customer inquiries. It covers both inquiries about a known Product or
departure and open-ended requests for custom travel. A qualified inquiry may
become a Proposal, enter a customer-completed Booking Session, or become a
staff-created Booking through the existing owner command for that target.

---

## 1. Summary

Voyant currently has useful but disconnected inquiry primitives:

- Storefront lead intake creates a Person and a `customer_signal` with kind
  `inquiry` or `request_offer`.
- Product checkout can create a separate `booking_inquiry`.
- Customer signals carry assignment and follow-up state, but have no packaged
  Inquiry queue or qualification command.
- Proposals provide the correct sales pipeline after a request is qualified.
- A Proposals runtime can create an inquiry-shaped Proposal, but there is no
  tracked in-repository consumer and no durable source inquiry handoff.

The result is capture without a complete operating workflow. The same customer
request can land in different tables depending on which form they used, and
staff cannot reliably answer these questions:

- What is new, unassigned, overdue, or waiting for a customer?
- Is this request for a particular Product, a departure, or something custom?
- Who owns the next action and when is it due?
- What conversations and changes have happened?
- Was it declined, lost, duplicated, converted to a Proposal, or booked?
- Can we prove which inquiry produced the Proposal or Booking?

This RFC introduces a first-class **Inquiry** owned by Relationships, backed by
one lifecycle and one operator workspace. It retires inquiry use of
`customer_signals` and `booking_inquiries` after a data-preserving migration.

The canonical flow becomes:

```text
customer request
      |
      v
   Inquiry  --triage / assign / contact / gather requirements--+
      |                                                       |
      +--> closed: lost / spam / duplicate / not serviceable  |
      |                                                       |
[...]

PR #4838, opened 2026-08-17
Build first-class inquiry management

mihaipxm:
Tracks #4837.

This is the single consolidated Inquiry-management PR. It includes the complete stacked implementation and the latest `main` merge, so review and merge happen through one conflict-resolved branch.

## Scope

- accepted Inquiry architecture and canonical terminology; `Inbox` remains reserved for Conversations
- Relationships-owned Inquiry aggregate, lifecycle, assignments, SLA/first-response handling, close/reopen, saved views, stable pagination, activities/timeline, target links, and transactional outbox events
- guarded public intake with immutable owner-validated target snapshots and replay-safe provenance
- durable Inquiry → Proposal and Inquiry → Booking Session conversions with operation-scoped idempotency, owner events, provenance, replay, refusals, and operator actions
- EN/RO operator queue and workspace with pagination, accessible lifecycle actions, activities, target management, conversion actions, attachments, and privacy controls
- private Inquiry attachments through Media-owned document storage, authenticated delivery, durable cleanup, export, and controlled erasure
- notifications, realtime invalidation, SLA scanning, reporting projections, and semantic admin destinations
- canonical legacy cutover for Customer Signals, Booking Inquiries, and the public Leads compatibility surface with resumable replay-safe adoption
- Inquiry MCP/Tools for reads, workflow writes, activities, targets, and Booking Session handoff through the same authenticated services
- canonical contracts, timestamped Drizzle migrations/snapshots, OpenAPI documents, generated clients, graph declarations, architecture ratchets, and changesets

## Consolidated stacked PRs

- #4839, #4840, #4842, #4844, #4846
- #4857, #4858, #4859
- #4862, #4864, #4867

## Verification

- full local workspace build: 117/117 tasks successful
- production Operator build: 20 initial JS preloads, within the packaged acceptance budget
- fresh Docker PostgreSQL database: all 32 generated migration/setup entries applied with zero failures; second-generation migration checks reported no drift
- exact local Operator runtime image booted successfully and `/healthz` returned `ok`
- demo seed completed on the disposable database
- authenticated Docker API smoke: Inquiry create 201, target add 201, detail read 200 with the canonical Product target projection
- Chrome reached the exact Docker sign-in surface without console errors; the remaining UI pass is being completed against the same container
- focused package typechecks/tests, graph/import-cheap, OpenAPI, migration, event authority, agent-tool/write coverage, Biome, and diff checks are green across the consolidated stack

[...]

PR #4867, opened 2026-08-18
feat(inquiries): cut over legacy inquiry capture

mihaipxm:
Part of #4837.

Moves new Booking Inquiry and /leads writes through canonical Relationships Inquiry commands while retaining compatibility response shapes and historical reads. Adds a bounded, resumable legacy adopter for Customer Signals and Booking Inquiries with replay-safe provenance, owner-validated target materialization, pending reconciliation, and canonical conversion provenance. Retires the unused Proposals checkout-inquiry seam and prevents retired Customer Signal write kinds from returning.

Verification:
- Drizzle generation and second-pass no-diff
- relationships-contracts, bookings, proposals-contracts, and public-api-client typechecks passed
- focused tests: 22 passed, 12 DB-gated skipped
- focused Relationships legacy source probe passed
- graph, retired-surface, symbol, Relationships runtime, migration boundary/cutline/identity checks passed
- Relationships and Public API full typechecks were attempted with 8GB and stopped after prolonged silent compilation; no diagnostics before stop
- runtime-ports aggregate currently exposes a pre-existing false-positive on the already-integrated Inquiry SLA config read; tracked for the parent reconciliation

No legacy table or compatibility read route is deleted. Destructive retirement remains gated on reconciliation reports reaching zero.


mihaipxm:
Consolidated into #4838 via feat/inquiry-management at ee1092fc8a after independent review approval. Closing this stacked PR so the remaining review and merge happen on the single parent PR.

PR #4930, opened 2026-08-27
Redesign the inquiry queue and detail workspace

mihaipxm:
## What

The inquiry detail page was a column of permanently-open forms — two conversion
forms below the timeline, a "Work details" form under them, and closing wired to
a select sitting next to a red button in the sidebar. Nothing told you what the
record was without scrolling, and every panel carried unsaved state you could not
see from the top.

It now reads as a record with actions:

- **Header** — subject, status/priority/kind, and the one or two things this
  status allows. Conversions are primary actions here, not forms further down.
  Everything else lives behind a "More actions" menu.
- **Main column** — Overview / Activity / Attachments tabs, so the page stops
  being an endless scroll.
- **Rail** — assignment, follow-up, contact, where it came from, and (once
  resolved) the outcome.

This is the header + tabs + rail shape `PersonDetailPage` and
`OrganizationDetailPage` already use. The queue picks up the search /
filter-popover / skeleton-rows / empty-state pattern from `PersonList`.

## Against the brief

1. **Dialogs and sheets instead of inline forms.** Notes + follow-up (sheet),
   close with outcome, create proposal, start booking, record activity, add file,
   link a product, and leave-unassigned-with-a-reason are each their own dialog.
2. **Library components instead of raw HTML.** `datetime-local` → `DateTimePicker`,
   `<input type="checkbox">` → `Checkbox`, hand-rolled view chips → `SegmentedControl`,
   hand-rolled pager buttons → `Button`, bare "no rows" text → `Empty`, "Loading…"
   row → `Skeleton` rows.
3. **Polish.** Status and priority badges are tiered by meaning (the same tone
   scale the bookings tables use); "Normal"/"Low" priority renders as plain text
   rather than a badge on every row, so the column only speaks up when it matters.
4. **Copy.** Rewritten for non-technical staff in English and Romanian, with every
   unused string removed. `triaged` reads as "Reviewed" and the action as "Mark as
   reviewed"; the stored enum value is untouched.
5. **Links to references.** The linked person and organization, `mailto:`/`tel:`
   for the contact, the product a target points at, the public page the customer
   submitted from, and the original of a duplicate close.

A destination the host has not bound resolves to `"#"` through `useAdminHref`,
which as an `href` looks like a working link and goes nowhere — the host now reads
that as "no link" and the page renders the plain value instead.

## Bugs fixed on the way through

- The queue's default `actionable` view had **no chip**, so the page opened with
  every saved view unselected and no way back to the default.
- `onRecordFirstResponse` was accepted as a prop and **never rendered**, so the
  first response could not be stamped from the UI at all.
- `InquiryOwnerField` seeded its display with `useState` and kept showing whoever
  owned the inquiry when it mounted, even after the owner changed underneath it.
- A qualified inquiry with no linked product disabled **both** conversions and
  said nothing about why — the same "a disabled button has to say why" lesson
  #4838 recorded for the lifecycle buttons.
- Trip-brief dates used the short numeric format (`9/14/2026`) while every other
  date on the page used the medium one (`Aug 18, 2026`).
- Clearing the owner required a reason typed into an always-visible box that meant
  nothing until you happened to clear the owner afterwards; it is now the dialog
  the API's own refusal implies.

## Breaking

`InquiryWorkspaceProps.onBack` is removed. The admin chrome owns the breadcrumb
trail — the host now registers it with `useAdminBreadcrumbs`, as the person and
organization hosts already do. No in-repo consumer passed it.

`@voyant-travel/relationships-react` is 0.x and this is a minor bump.

## Verification

- `lint`, `typecheck`, `test` (71 passing), and `build` green for the package
- `pnpm verify:architecture` green end to end in one uninterrupted run, with the
  working tree unchanged around it
- Rendered every state (queue, qualified, new/unassigned, closed) through a
  throwaway Vite harness against the real stylesheet, in light and dark and in
  both locales, and clicked through each dialog. The harness is not committed.

The added declarative rule is one `AdminDestinations` member (`product.detail`,
shape-locked against the three packages that already declare it) — no new
`scripts/check-*.mjs`.


mihaipxm:
Addressed all three, and the P1 turned out to reach further than the comment said.

**P1 — reopen could not carry the reason.** Confirmed against the service: `reopenInquiry` accepts `unassignedReason` and is the *only* command that can set one on a closed record, because `assignInquiry` throws `INQUIRY_ALREADY_RESOLVED` for closed/converted (`packages/relationships/src/service/inquiries.ts:1324-1330`). So the sidebar escape hatch was indeed dead, and gating Reopen on a reason nobody could enter stranded the inquiry permanently.

Reopen now opens `InquiryReopenDialog`, which collects the reason when the record has neither owner nor stored reason, and the button is never disabled. It also offers the follow-up date, because `reopenInquiry` writes `nextActionAt` from the command unconditionally — omitting it clears whatever was stored and triage then blocks on the field that was just wiped.

The same service guard means the assignment controls were live on a resolved inquiry where all three writes are guaranteed 400s. They are now disabled with "A closed inquiry cannot be reassigned. Reopen it first."

**P2 — waiver committed on toggle.** Right. `noFollowUpExpected` is now local to the sheet and handed back only after a successful save, so a cancelled edit cannot enable the lifecycle button or send a waiver the operator took back.

**P2 — unassign closed before awaiting.** Right, and the work-details sheet and attachment upload had the same shape. All three now await, keep the dialog open on rejection, and show an error.

Verified the reopen path end to end against the real components: the command carries `{"unassignedReason":"…","nextActionAt":null}`, which is what satisfies `INQUIRY_ASSIGNMENT_REQUIRED`. Package lint/typecheck/test (72)/build green, and `pnpm verify:architecture` green end to end with the tree unchanged around it.

PR #4943, opened 2026-08-27
Search payments by name and booking number, and show the booking

mihaipxm:
## What

The payments index searched a payment's own reference number and nothing else — the one identifier staff rarely have in hand. A caller gives a name and quotes a booking number, and neither found anything.

`search` now spans exactly what the row shows: its reference, the invoice or booking it settles, and the party it is with — matched case-insensitively on partials, on both sides of the customer/supplier union.

| side | search now matches |
|---|---|
| customer | `payments.reference_number`, `invoices.invoice_number`, `bookings.booking_number`, person first/last/full name, organization name |
| supplier | `supplier_payments.reference_number`, `bookings.booking_number`, supplier name |

Party name is included because it is the column the operator is reading — "search by person name" on a row whose Party cell says `Vacante Corporate SRL` should find it by either.

## The booking column

The list gains a **Booking** column. To fill it, the customer half of the UNION stopped hardcoding `booking_id` / `booking_number` to `NULL` and resolves them through the invoice it settles (`LEFT JOIN bookings b ON b.id = i.booking_id`). `getPaymentById` got the same join, because it is documented as returning the same enriched shape — and a row that claims to be the same shape must carry the same values. A side effect worth naming: a **customer** payment now gets its "View booking" link on the payment detail page, which was previously dead for that kind.

The Reference cell no longer doubles as the booking: it shows the invoice number for a customer payment, the payment's own reference for a supplier one, and only carries the reference as a sub-line when there is a document number above it.

## Verification

The projection is a UNION with different joins per side, so what a term reaches is a property of the SQL — a fixture test would only prove the compiler works. `packages/finance/tests/integration/payments-list-search.test.ts` drives `listAllPayments` over real Postgres across a seeded person, organization, two bookings and a supplier payment, and is **registered in the `db-integration` lane by name** (that lane lists files individually, so an unregistered integration test runs nowhere).

Local, on this branch:

- `packages/finance` typecheck — `exit=0`, `0` × `error TS`
- `packages/finance-react` typecheck — `exit=0`, `0` × `error TS`
- `@voyant-travel/finance` test — 90 passed / 25 skipped (115 files), 723 passed
- `@voyant-travel/finance-react` test — 22 files, 70 passed
- `biome check` on all 8 changed files — clean
- `pnpm verify:openapi-drift` — `exit=0`, 146 documents match

`packages/finance/openapi/admin/finance.json` and `packages/admin-api-client/src/generated/finance.ts` are regenerated: the only diff in each is the new `search` description.

The integration test itself was **not** run on this machine — the local Docker engine is down, so it was dispatched to lab1 against a real Postgres 16 with the operator migrations applied. Results are posted as a follow-up comment.

## Notes for review

- No schema change: `unifiedPaymentSchema` already declared `bookingId` / `bookingNumber` nullable; the customer side was simply never populating them.
- A whitespace-only `search` is now treated as no search rather than an `ILIKE '%   %'` that matches nothing.


mihaipxm:
## Integration results — real Postgres

The local Docker engine is down on my machine, and lab1 is offline (node key expired, last seen 7 days), so this ran on disposable Sprites with Postgres installed natively, `pnpm install --frozen-lockfile`, and `operator db:migrate` (33 migrations, 495 tables) — the same bootstrap the `db-integration` lane uses.

| suite | REAL_EXIT | summary |
|---|---|---|
| `payments-list-search.test.ts` | **0** | 1 file passed, **11 tests passed** |
| `routes.test.ts` | 1 | 5 failed / 148 passed (153) |
| `profitability.test.ts` | **0** | 1 file passed, 18 tests passed |

**The new suite genuinely executed** — 11 tests ran, not skipped. Worth stating explicitly because the suite is `describe.skipIf(!process.env.TEST_DATABASE_URL)`, so a missing env var would have reported a cheerful green while checking nothing.

### The `routes.test.ts` failures are pre-existing

A control run on the base commit `dbe606a543`, in an identical environment, produced **byte-identical** results: same `5 failed | 148 passed`, same three named failures. The diff is exonerated.

They sit in `Invoice from booking` and `Booking Payment Schedules` — paths this change does not touch. One is an ordering-sensitive string assertion:

```
Expected: "... 1 × Cameră single + 1 × DBL | ..."
Received: "... 1 × DBL + 1 × Cameră single | ..."
```

The other two are `expected 500 to be 201` on default deposit/balance payment plans.

**Caveat worth carrying forward:** the Sprite's apt default installed **PostgreSQL 18.6, not 16**, which is a plausible cause of an unstable-ordering failure. So I can say these 5 are identical with and without this change; I cannot say whether they also fail in CI's PG16 lane. That is a separate thread from this PR either way.

### Gotcha for anyone re-running this

Use `localhost`, not `127.0.0.1`, in `TEST_DATABASE_URL`. `packages/finance/tests/integration/routes.test.ts:84` rewrites a `127.0.0.1` URL to a `voyant_finance_test` database for isolation, which silently produces `153 skipped` and exit 1 — a failure that looks nothing like its cause.

mihaipxm:
## Full architecture chain — green

`pnpm verify:architecture`, one uninterrupted run on this branch:

```
REAL_EXIT=0  elapsed=414s
TREE UNCHANGED
```

2040 lines, zero `ELIFECYCLE`, terminating on the last link (`combobox-filtering: OK`). Stated precisely because the chain short-circuits on its first failure, so a partial run is a false green — and my first two attempts at this were killed by the runner mid-chain and proved nothing in either direction. `git status --porcelain` was diffed before and after, since checkers in this chain can mutate the tree.

Also added since the PR was opened: `packages/finance-react/src/payments-page.test.tsx` — 5 tests rendering the page over a seeded query cache, pinning that both payment kinds show their booking in the new column and that the Reference cell stays on the document instead of doubling the booking number. One of them asserts the booking number appears exactly once, which is also what proves the rows actually rendered rather than the skeleton.

`@voyant-travel/finance-react`: 23 files, 75 passed.

